PR #4, opened 2023-12-17
Fix docs regression & add show in playground

sasial-dev:


sasial-dev:
I've extensively tested this locally and am happy with stability.

Issue #7, opened 2023-12-17
alloc can reassign `outgoing_buff` which causes the write call to write to the old buffer


jackdotink:
title explains it

Issue #6, opened 2023-12-17
fire except doesn't generate closing end after the firing for loop


jackdotink:
title explains itself

PR #13, opened 2023-12-18
irgen and output rewrite - actual maintainable code!

### LINKED ISSUE #5: syntax error when generating event listener
When zap generates the reliable/unreliable event listener it uses `if` for an event with id 1, and uses `elseif` for all others. This is a syntax error if the event with id 1 isn't being listened to in that single listener.

jackdotink:
For *far* too long have we suffered with the unmaintainable code of the `irgen` and `output` modules. No more! With this full rewrite we will be able to actually maintain these modules into the future.

This closes #5 #6 and #7.


zap/src/output/client.rs
```diff
@@ -1,260 +1,363 @@
...
+	fn dedent(&mut self) {
+		self.tabs -= 1;
+	}
+
+	fn push_indent(&mut self) {
+		for _ in 0..self.tabs {
+			self.push("\t");
+		}
+	}
+}
```
sasial-dev: These methods should be inherited from the trait.

zap/src/output/client.rs
```diff
@@ -1,260 +1,363 @@
...
+			self.push("if ");
+		} else {
+			self.push("elseif ");
 		}
 
-		writeln!(f, "--!native")?;
-		writeln!(f, "--!optimize 2")?;
-		writeln!(f, "-- Generated by zap v{}", env!("CARGO_PKG_VERSION"))?;
+		self.push(&format!("id == {id} then"));
+		self.push("\n");
```
sasial-dev, line 99: Shouldn't this just be:
```suggestion
		self.push_line(&format!("id == {id} then"));
```

zap/src/output/client.rs
```diff
@@ -1,260 +1,363 @@
...
+	fn push_reliable_footer(&mut self) {
+		self.push_line("else");
+		self.indent();
+		self.push_line("error(\"Unknown event id\")");
+		self.dedent();
+		self.push_line("end");
+		self.dedent();
+		self.push_line("end");
+		self.dedent();
+		self.push_line("end)");
```
sasial-dev, line 125: Maybe this should use `self.tabs and from that decide how many "ends" it should insert?

zap/src/output/client.rs
```diff
@@ -1,260 +1,363 @@
...
+	}
+
+	fn push_reliable(&mut self) {
+		self.push_reliable_header();
+
+		let mut first = true;
+
+		for (i, ev) in self
```
sasial-dev, line 133: You shouldn't need the `first` var, you can just check the value of `i` instead.

zap/src/output/client.rs
```diff
@@ -1,260 +1,363 @@
...
+		self.dedent();
+		self.push_line("end)");
+	}
+
+	fn push_unreliable(&mut self) {
+		self.push_unreliable_header();
+
+		let mut first = true;
+
+		for (i, ev) in self
```
sasial-dev, line 205: Same here

zap/src/output/client.rs
```diff
@@ -1,260 +1,363 @@
...
+
+		self.push_callback_lists();
+
+		self.push_return();
+	}
 
-		Ok(())
+	pub fn into_inner(self) -> String {
+		self.buff
 	}
```
sasial-dev, line 375: Do you really need this function?

zap/src/output/mod.rs
```diff
@@ -1,40 +1,87 @@
...
-impl Display for TyDecl {
-	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-		let name = &self.name;
-		let ty = &self.ty;
-
-		writeln!(f, "export type {name} = {ty}")?;
+pub trait Output {
+	fn push(&mut self, s: &str);
+	fn indent(&mut self);
+	fn dedent(&mut self);
```
sasial-dev, line 11: `dedent` feels very unclear as a function name

zap/src/output/mod.rs
```diff
@@ -1,40 +1,87 @@
...
+	fn push_indent(&mut self);
 
-		writeln!(f, "function types.write_{name}(value: {name})")?;
+	fn push_line(&mut self, s: &str) {
+		self.push_indent();
+		self.push(s);
+		self.push("\n");
+	}
```
sasial-dev, line 18: This is the wrong order. It should go:
- Newline
- Push Indent
- Push String

to make it consistant with other langs iirc

zap/src/output/mod.rs
```diff
@@ -1,40 +1,87 @@
...
+			Stmt::Local(name, expr) => {
+				if let Some(expr) = expr {
+					self.push_line(&format!("local {name} = {expr}"));
+				} else {
+					self.push_line(&format!("local {name}"));
+				}
+			}
+
+			Stmt::Assign(var, expr) => self.push_line(&format!("{var} = {expr}")),
+			Stmt::Error(msg) => self.push_line(&format!("error({msg})")),
```
sasial-dev, line 35: Consider 'expanding the stack' by increasing the [error level](https://create.roblox.com/docs/reference/engine/globals/LuaGlobals#error) arg

zap/src/output/mod.rs
```diff
@@ -1,40 +1,87 @@
...
+			Stmt::Assert(cond, msg) => match msg {
+				Some(msg) => self.push_line(&format!("assert({cond}, {msg})")),
+				None => self.push_line(&format!("assert({cond})")),
+			},
+
+			Stmt::Call(var, method, args) => match method {
+				Some(method) => self.push_line(&format!(
+					"{var}:{method}({})",
```
sasial-dev, line 43: What about callbacks? `::Call` doesn't seem to account for stuff like `var.callback()`...

zap/src/output/server.rs
```diff
@@ -1,434 +1,514 @@
...
+	fn dedent(&mut self) {
+		self.tabs -= 1;
+	}
+
+	fn push_indent(&mut self) {
+		for _ in 0..self.tabs {
+			self.push("\t");
 		}
+	}
+}
```
sasial-dev: Same feedback as for the client

zap/src/output/server.rs
```diff
@@ -1,434 +1,514 @@
...
+		self.dedent();
+		self.push_line("end)");
+	}
+
+	fn push_reliable(&mut self) {
+		self.push_reliable_header();
+
+		let mut first = true;
+
+		for (i, ev) in self
```
sasial-dev, line 138: Same as with client, you shouldn't need `first` just use `i`.

zap/src/output/server.rs
```diff
@@ -1,434 +1,514 @@
...
+				self.push_line(&format!("unreliable:FireClient({player}, buff, outgoing_inst)"));
+			}
+		}
 
-		for (i, ev_decl) in file
+		self.dedent();
+		self.push_line("end,");
+	}
+
+	fn push_return_fire_all(&mut self, ev: &EvDecl, id: usize) {
```
sasial-dev, line 271: I wonder if `push_return_fire_*` should inherit from a base fuction to save some duplication

zap/src/output/server.rs
```diff
@@ -1,434 +1,514 @@
...
+
+		self.push_callback_lists();
+
+		self.push_return();
+	}
 
-		Ok(())
+	pub fn into_inner(self) -> String {
+		self.buff
 	}
```
sasial-dev, line 522: Same as before

zap/src/parser/mod.rs
```diff
@@ -114,17 +114,11 @@ impl Ty {
...
 			Ty::U32(_) => Some(4),
 
-			Ty::Str { len } => {
-				if len.is_exact() {
-					Some(len.min().unwrap() as usize)
-				} else {
-					None
-				}
-			}
+			Ty::Str { len } => len.exact().map(|len| len as usize),
```
sasial-dev, line 117: `.map` feels like a waste for a type assertion - can you assert w/o mapping?

sasial-dev (changes requested):
Some initial comments

zap/src/output/client.rs
jackdotink: this isn't possible in rust as traits cannot declare fields

zap/src/output/client.rs
jackdotink: `push_line` also pushes tabs, so we can't do this. We could move `\n` to the same push though.

zap/src/output/client.rs
jackdotink: It's always going to be 3 because of the above code - general rule is that when you explicitly open a scope you must explicitly close it somewhere else.

zap/src/output/client.rs
jackdotink: The value of `i` relates to the id of the event, and is not guaranteed to be iterated over - this is the cause of #5 in the last output generator.

zap/src/output/client.rs
jackdotink: same as above

zap/src/output/client.rs
jackdotink: Not really, we can remove it and put it into `output`

zap/src/output/mod.rs
jackdotink: It's the only thing I could think of that's two letters (please don't ask me to make it outdent :c). As a side not python also uses it?

zap/src/output/mod.rs
jackdotink: This is actually correct order, we start on an empty line, push line content, and then push a newline to the end of the line. This is why files after being formatted end with an empty line, the last line with content is suffixed with a newline.

zap/src/output/client.rs
sasial-dev: Seems like it is https://doc.rust-lang.org/book/ch10-02-traits.html ?

zap/src/output/client.rs
sasial-dev: Let's keep them explicitly seperate and add a comment as to why we aren't using `push_line`.

zap/src/output/client.rs
sasial-dev: Sounds good - I agree.

zap/src/output/mod.rs
jackdotink: Errors are thrown in serdes failures. If des fails the fault lies with zap, and we want as detailed of an error as possible. If ser fails the fault lies with the user, and they typically *also* want a detailed error so they know what they did wrong.

zap/src/output/mod.rs
jackdotink: the `Var` enum accounts for indexing separately.

zap/src/output/server.rs
jackdotink: same as server

zap/src/output/server.rs
jackdotink: same as server

zap/src/output/client.rs
sasial-dev: Got it.

zap/src/output/server.rs
jackdotink: I don't know if this is possible in rust, if it is I'll happily look into it.

zap/src/output/server.rs
jackdotink: same as before

zap/src/output/mod.rs
sasial-dev: That's not what I mean.
I meant: `error(msg)` -> `error(msg, 3)` or something like that.

zap/src/parser/mod.rs
sasial-dev: This is mapping an `Option`, my bad.

jackdotink: `len.exact` returns an `Option<T>`

zap/src/output/server.rs
sasial-dev: Should've been clearer, by inheriting I mean just calling a base one in each of these

zap/src/output/mod.rs
sasial-dev: `rmindent`? I'm not sure how much I like that either though...
maybe `dedent` is the best?

sasial-dev: Let's keep it as `dedent`.

zap/src/irgen/gen.rs
```diff
@@ -1,517 +1,603 @@
...
 
-					self.block_end();
+		Ty::I8(..) => buffer_writei8(&mut stmts, from_expr),
+		Ty::I16(..) => buffer_writei16(&mut stmts, from_expr),
+		Ty::I32(..) => buffer_writei32(&mut stmts, from_expr),
 
-					self.block_end();
+		Ty::Str { len } => {
+			if let Some(len) = len.exact_f64() {
+				if gen_checks {
```
sasial-dev, line 280: This should be moved up scope to line 260 like the rest of the `gen_checks`.

jackdotink: I'd rather keep the string checks grouped together as it might cause readers of the code to miss the check for exact strings if it were moved above.

zap/src/irgen/gen.rs
```diff
@@ -1,517 +1,603 @@
...
+				to.clone(),
+				Var::from("incoming_inst").eindex(buffer_readu16()).into(),
+			);
 
-				self.assign(
-					into.clone(),
-					Expr::Vector3(Box::new("x".into()), Box::new("y".into()), Box::new("z".into())),
+			// Assert that the instance is not nil even if we don't want checks
+			// because roblox cannot ensure the instance's existance at the destination
+			assert(&mut stmts, Expr::from(to.clone()).neq(Expr::Nil), None);
```
sasial-dev, line 541: Is this the right way to go? As an error will... error when we might want to handle gracefully?

zap/src/irgen/gen.rs
```diff
@@ -1,517 +1,606 @@
...
+				Var::from("incoming_inst").eindex(buffer_readu16()).into(),
+			);
 
-				self.read_num("x".into(), NumTy::F32);
-				self.read_num("y".into(), NumTy::F32);
-				self.read_num("z".into(), NumTy::F32);
+			// Assert that the instance is not nil even if we don't want checks
+			// because roblox cannot ensure the instance's existance at the destination.
+			// Only do this for strict instances, optional instances can be nil.
+			if *strict {
```
sasial-dev, line 542: You shouldn't need to clone to check if it's a strict? Feels like a perfect case for a borrow...

sasial-dev (approved):
It's time.

PR #15, opened 2023-12-18
Add support for seperate output file locations

### LINKED ISSUE #14: [FEAT] Seperate Output File Locations
## Describe your feature
Currently, zap outputs in a folder with the client and shared code together. When using rojo, it can become very painful to have server and client code in the same directory. Zap should instead specifiy in the config file where the outputs should be. 

## Alternatives
- Stick with the extisting `network` folder for output.
- Define the outputs through the CLI instead of the config file

sasial-dev:
Closes #14.


jackdotink (approved):
lgtm

PR #16, opened 2023-12-19
Make ranges always max inclusive

### LINKED ISSUE #9: [FEAT] make range max inclusive by default
## Describe your feature
Rust uses the range syntax of `1..=5` because they have arrays/vecs which start indexing at 0, and the highest index of an array is one less than the length. In luau indexing starts at 1, and the highest index of an array *is* the length, so it would make more sense and be more intuitive to make the max part of a range inclusive.

This is a breaking change. It would change the behavior of any currently non-inclusive ranges, and cause parse errors for any inclusive ranges. As we are in early 0.x, this is acceptable for the benefit it will bring over time.

## Alternatives
* Using mathematical interval notation: `[0, 10)` = inclusive 0, exclusive 10
This would likely create more confusion, and it becomes harder to provide good syntax highlighting and other editor features.

* Using a function-like syntax: `Limit(u8, 0, 9)` = inclusive 0, inclusive 9
This has the change of making max inclusive, but creates large syntax that only works for numeric types.

* Using commas instead of `..`: `1, 9`
This would not allow for one sided ranges or exact ranges.

## Implementation Details
This makes the parsing and implementation for Ranges simpler, and should have minimal effect on Range usage.

## Additional context
N/A

jackdotink:
This change removes the syntax `1..=10` and makes `1..10` include 10.

closes #9

### DIFF
--- a/docs/config/types.md
+++ b/docs/config/types.md
@@ -44,10 +44,6 @@ Suppose we are trying to send the health of a player over the remote. We may def
 
 <CodeBlock code="type Health = u8 (0..100)" />
 
-Although, the type assertions will fail if 100 is passed. This is because **only the lower limit is inclusive**. To fix this, we must append a `=` to the end of the upper bound such as
-
-<CodeBlock code="type Data = u8 (0..=100)" />
-
 ## Strings
 
 Unlike numbers, strings do not have a maximum length across different types. They can be any length unless they are constrained.
--- a/docs/config/types.md
+++ b/docs/config/types.md
@@ -44,6 +44,7 @@ Suppose we are trying to send the health of a player over the remote. We may def
 
 <CodeBlock code="type Health = u8 (0..100)" />
 
+This works because limits are **inclusive**, and will include 0 and 100.
 ## Strings
 
 Unlike numbers, strings do not have a maximum length across different types. They can be any length unless they are constrained.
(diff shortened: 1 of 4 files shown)


docs/config/types.md
```diff
@@ -44,10 +44,6 @@ Suppose we are trying to send the health of a player over the remote. We may def
 
 <CodeBlock code="type Health = u8 (0..100)" />
```
sasial-dev, line 46: Might want to add an explaination as to _why_ it works
```suggestion

This works because limits are **inclusive**, and will include 0 and 100.
```

zap/src/util.rs
```diff
@@ -77,32 +51,16 @@ impl<T: Num + NumCast + Copy + Display + PrimInt> Range<T> {
 
 impl<T: Num + NumCast + Copy + Display> Default for Range<T> {
 	fn default() -> Self {
```
sasial-dev, line 53: Rust idom question: Are you actually using `.default` anywhere for Range?

jackdotink: Pretty sure we're using `unwrap_or_default` somewhere, but I could be mistaken

sasial-dev: Yes you are - no worries then

PR #31, opened 2023-12-29
remove local spam

jackdotink:
At the moment Zap generated code is at risk of hitting the local limit. This change removes the creation of a local for each write and instead uses a single local.


sasial-dev (approved):
LGTM!

PR #33, opened 2023-12-30
remove instance overhead

### LINKED ISSUE #32: [FEAT] Remove Instance Overhead
## Describe your feature
Currently instances are placed into a single large array, and then the index of the instance in that array is written to the buffer. This wastes the two bytes written to the buffer. As Zap serializes and deserializes in the same order, instances will be written to the array in the same order.

A type like this: `Instance(BasePart)` is currently outputted as the following:
```lua
function types.write_basePart(value: basePart)
	local pos = alloc(2)
	buffer.writeu16(outgoing_buff, pos, alloc_inst(value))
end
function types.read_basePart()
	local value;
	value = incoming_inst[buffer.readu16(incoming_buff, read(2))]
	assert(value ~= nil)
	return value
end
```
With this change implemented this will be changed to this:
```lua
function types.write_basePart(value: basePart)
	table.insert(outgoing_inst, value)
end
function types.read_basePart()
	local value;
	value = table.remove(incoming_inst)
	assert(value ~= nil)
	return value
end
```
Again, this is possible because we serialize and deserialize in the same order - so instances will be written and read from the `inst` buffer in the same order.

## Alternatives
I have not considered any alternatives at this moment.

## Implementation Details
This could cause issues if we ever have non-deterministic serialization or deserialization order in the future.

## Additional context
N/A

jackdotink:
This also fixes instances not being sent when `FireAll`, `FireExcept` and `FireList` are used.

Closes #32

### DIFF
--- a/zap/src/irgen/des.rs
+++ b/zap/src/irgen/des.rs
@@ -165,7 +165,13 @@ impl Des {
 
 				match **ty {
 					Ty::Instance(class) => {
-						self.push_assign(into.clone(), Var::from("incoming_inst").eindex(self.readu16()).into());
+						self.push_assign(Var::from("incoming_ipos"), Expr::from("incoming_ipos").add(1.0.into()));
+						self.push_assign(
+							into.clone(),
+							Var::from("incoming_inst")
+								.eindex(Var::from("incoming_ipos").into())
+								.into(),
+						);
 
 						if self.checks && class.is_some() {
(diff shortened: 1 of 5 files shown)


jackdotink:
closes #32

sasial-dev (approved):
I've tested the output - LGTM.

PR #34, opened 2023-12-31
Delay event connection 5 frames on client

nezuo:
This change is important to support events that are fired the frame the client joins the game. Without the delay, the client won't have enough time to connect to the events.


sasial-dev:
Closed after heavy discussion in favour of the preferred solution of implementing a queue.

PR #40, opened 2023-12-31
Don't create remotes if they already exist

nezuo:


zap/src/output/luau/client.luau
```diff
@@ -1,15 +1,18 @@
...
 	error("Cannot use the client module on the server!")
 end
 
-local reliable = game:GetService("ReplicatedStorage"):WaitForChild("ZAP_RELIABLE")
-local unreliable = game:GetService("ReplicatedStorage"):WaitForChild("ZAP_UNRELIABLE")
+local reliable = ReplicatedStorage:WaitForChild("ZAP_RELIABLE")
+local unreliable = ReplicatedStorage:WaitForChild("ZAP_UNRELIABLE")
+
+assert(reliable:IsA("RemoteEvent"), "Expected ZAP_RELIABLE to be a RemoteEvent")
+assert(unreliable:IsA("UnreliableRemoteEvent"), "Expected ZAP_UNRELIABLE to be an UnreliableRemoteEvent")
```
sasial-dev, line 9: I know I just told you to assert, but as the server created it - is it safer to just assert here as we know it's from the server? Or, should we keep the error to protect from an accident?

```suggestion
local reliable = ReplicatedStorage:WaitForChild("ZAP_RELIABLE") :: RemoteEvent
local unreliable = ReplicatedStorage:WaitForChild("ZAP_UNRELIABLE") :: UnreliableRemoteEvent
```

nezuo: Since this lets the user create the remotes, I think we should keep the runtime asserts. For example, a user could accidentally use a `RemoteEvent` instead of an `UnreliableRemoteEvent`.

sasial-dev: Sounds good!

sasial-dev (approved):
LGTM! Thank you for going through with the extra things.

PR #45, opened 2024-01-01
Fix fire in player added

nezuo:
Firing a remote on the server inside `PlayerAdded` is not guaranteed to exist because the user's connection can run before zap's. This fixes the issue by initializing the player inside of fire functions instead of `PlayerAdded`.


zap/src/output/luau/client.rs
```diff
@@ -89,6 +89,7 @@ impl<'src> ClientOutput<'src> {
 			self.config.event_id_ty().size()
 		));
 	}
+
```
sasial-dev, line 92: I don't think this diff is worth it tbh for this file. I'm happy to keep the formatting in `base.luau` though

zap/src/output/luau/server.luau
```diff
@@ -20,14 +20,18 @@ end
...
+local function load_player(player: Player)
+	if player_map[player] then
+		load(player_map[player])
+	else
+		load({
+			buff = buffer.create(64),
+			used = 0,
+			size = 64,
+			inst = {},
+		})
```
jackdotink, line 32: Replace with `load_empty()`

jackdotink (approved):

sasial-dev (approved):

PR #51, opened 2024-01-03
rewrite convert

jackdotink:
This PR rewrites all of `convert.rs` and makes some modifications elsewhere. This allows us to properly check unbounded recursive types, get the minimum and maximum size of types for unreliable size checks, and will be easier to work with in the future.

### DIFF
--- a/zap/src/config.rs
+++ b/zap/src/config.rs
@@ -98,6 +98,11 @@ pub enum Ty<'src> {
 }
 
 impl<'src> Ty<'src> {
+	/// Returns the amount of data used by this type in bytes.
+	///
+	/// Note that this is not the same as the size of the type in the buffer.
+	/// For example, an `Instance` will always send 4 bytes of data, but the
+	/// size of the type in the buffer will be 0 bytes.
 	pub fn size(
 		&self,
 		tydecls: &HashMap<&'src str, TyDecl<'src>>,
@@ -169,14 +174,14 @@ impl<'src> Ty<'src> {
 			Self::Enum(enum_ty) => enum_ty.size(tydecls, recursed),
 			Self::Struct(struct_ty) => struct_ty.size(tydecls, recursed),
 
-			Self::Instance(_) => (0, Some(0)),
+			Self::Instance(_) => (4, Some(4)),
 
 			Self::Boolean => (1, Some(1)),
 			Self::Color3 => (12, Some(12)),
 			Self::Vector3 => (12, Some(12)),
 			Self::AlignedCFrame => (13, Some(13)),
 			Self::CFrame => (48, Some(48)),
-			Self::Unknown => (0, Some(0)),
+			Self::Unknown => (0, None),
 		}
 	}
 }
--- a/zap/src/config.rs
+++ b/zap/src/config.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashSet, fmt::Display};
+use std::{
+	collections::{HashMap, HashSet},
+	fmt::Display,
+};
 
 #[derive(Debug, Clone)]
 pub struct Config<'src> {
@@ -8,8 +11,8 @@ pub struct Config<'src> {
 	pub write_checks: bool,
 	pub typescript: bool,
 
-	pub server_output: Option<&'src str>,
-	pub client_output: Option<&'src str>,
+	pub server_output: &'src str,
+	pub client_output: &'src str,
 
 	pub casing: Casing,
 }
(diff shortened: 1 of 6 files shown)


zap/src/config.rs
```diff
@@ -141,22 +192,41 @@ pub enum Enum<'src> {
...
-						size += ty_size;
+					if let Some(ty_max) = ty_max {
+						if let Some(current_max) = max {
+							if ty_max > current_max {
+								max = Some(ty_max);
+							}
+						}
 					} else {
-						return None;
+						max = None;
```
sasial-dev, line 225: Setting this to max may overwrite the previously defined Max type. Not sure if it's intended.

zap/src/parser/convert.rs
```diff
@@ -1,142 +1,97 @@
...
 
-		for decl in self.decls {
-			match decl {
-				SyntaxDecl::Ty(tydecl) => tydecls.push(tydecl.into_config(state)),
-				SyntaxDecl::Ev(evdecl) => evdecls.push(evdecl.into_config(state)),
+		let write_checks = self.boolean_opt("write_checks", true, &config.opts).0;
+		let typescript = self.boolean_opt("typescript", false, &config.opts).0;
+
+		let server_output = self.str_opt("server_output", "network/server.lua", &config.opts).0;
+		let client_output = self.str_opt("client_output", "network/client.lua", &config.opts).0;
```
sasial-dev, line 74:
```suggestion
		let (write_checks, _) = self.boolean_opt("write_checks", true, &config.opts);
		let (typescript, _) = self.boolean_opt("typescript", false, &config.opts);

		let (server_output, _) = self.str_opt("server_output", "network/server.lua", &config.opts);
		let (client_output, _) = self.str_opt("client_output", "network/client.lua", &config.opts);
```

zap/src/parser/convert.rs
```diff
@@ -166,228 +204,326 @@ impl<'src> SyntaxEvDecl<'src> {
...
+				}
+			}
+
+			SyntaxTyKind::Ref(ref_ty) => {
+				let ref_name = ref_ty.name;
+
+				match ref_name {
+					ref_name if ref_name == name => Some(*ref_ty),
+
+					"boolean" | "Color3" | "Vector3" | "AlignedCFrame" | "CFrame" | "unknown" => None,
```
sasial-dev, line 382: Should we put these somewhere else so we don't have to hardcode primatives in two places?

zap/src/parser/reports.rs
```diff
@@ -227,6 +235,15 @@ impl<'src> Report<'src> {
 			Self::AnalyzeInvalidOptionalType { span, .. } => {
 				vec![Label::primary((), span.clone()).with_message("must be removed")]
 			}
+
+			Self::AnalyzeUnboundedRecursiveType {
+				decl_span, use_span, ..
+			} => {
+				vec![
+					Label::primary((), decl_span.clone()).with_message("declared here"),
+					Label::secondary((), use_span.clone()).with_message("used recursively here"),
```
sasial-dev, line 244: Where it is used recursively should be the primary issue
```suggestion
					Label::secondary((), decl_span.clone()).with_message("declared here"),
					Label::primary((), use_span.clone()).with_message("used recursively here"),
```

zap/src/parser/reports.rs
```diff
@@ -244,7 +261,7 @@ impl<'src> Report<'src> {
 				format!("all unreliable events must be under {max_size} bytes in size"),
 				"consider adding a upper limit to any arrays or strings".to_string(),
 				"upper limits can be added for arrays by doing `[..10]`".to_string(),
-				"upper limits can be added for strings by doing `[..10]`".to_string(),
+				"upper limits can be added for strings by doing `(..10)`".to_string(),
```
sasial-dev, line 264: Maybe we should have this shows the upper limit syntax for the correct type

zap/src/parser/reports.rs
```diff
@@ -273,7 +290,13 @@ impl<'src> Report<'src> {
 				format!("(inclusive) max: {}", max),
 			]),
 			Self::AnalyzeInvalidOptionalType { .. } => Some(vec![
-				"this type can only be used when it is not marked as optional".to_string(),
+				"you cannot have 'double optional' types, where a type is optional twice".to_string(),
+				"maps cannot have optional keys or values, as those are impossible to represent in luau".to_string(),
+				"additionally the `unknown` type cannot be optional".to_string(),
```
sasial-dev, line 295: again we should use the props to show the 'correct' one here

sasial-dev (changes requested):
Looking pretty good

zap/src/parser/convert.rs
jackdotink: I'd like that, but idk what to do for it

zap/src/parser/reports.rs
jackdotink: potentially, but out of scope for this PR

zap/src/parser/reports.rs
jackdotink: also out of scope for this PR

zap/src/config.rs
```diff
@@ -95,37 +98,85 @@ pub enum Ty<'src> {
...
+					let tydecl = tydecls.get(name).unwrap();
+
+					tydecl.ty.size(tydecls, recursed)
 				}
 			}
+
+			Self::Enum(enum_ty) => enum_ty.size(tydecls, recursed),
+			Self::Struct(struct_ty) => struct_ty.size(tydecls, recursed),
+
+			Self::Instance(_) => (0, Some(0)),
```
sasial-dev, line 172: Should we add a comment explaining why it's `Some(0)` over `None` or do you think it's selfexplainatory?

zap/src/parser/convert.rs
```diff
@@ -1,142 +1,97 @@
...
 
-		for decl in self.decls {
-			match decl {
-				SyntaxDecl::Ty(tydecl) => tydecls.push(tydecl.into_config(state)),
-				SyntaxDecl::Ev(evdecl) => evdecls.push(evdecl.into_config(state)),
+		let (write_checks, _) = self.boolean_opt("write_checks", true, &config.opts);
+		let (typescript, _) = self.boolean_opt("typescript", false, &config.opts);
+
+		let (server_output, _) = self.str_opt("server_output", "network/server.lua", &config.opts);
+		let (client_output, _) = self.str_opt("client_output", "network/client.lua", &config.opts);
```
sasial-dev, line 74: Let's make this even more explicit

```suggestion
		let (write_checks, ..) = self.boolean_opt("write_checks", true, &config.opts);
		let (typescript, ..) = self.boolean_opt("typescript", false, &config.opts);

		let (server_output, ..) = self.str_opt("server_output", "network/server.lua", &config.opts);
		let (client_output, ..) = self.str_opt("client_output", "network/client.lua", &config.opts);
```

zap/src/parser/convert.rs
```diff
@@ -166,228 +204,326 @@ impl<'src> SyntaxEvDecl<'src> {
...
+					.map(|variant| {
+						if variant.1.fields.iter().any(|field| field.0.name == tag_name) {
+							self.report(Report::AnalyzeEnumTagUsed {
+								tag_span: tag.span(),
+								used_span: variant.0.span(),
+								tag: tag_name,
+							});
+						}
+
+						(variant.0.name, self.struct_ty(&variant.1))
```
sasial-dev, line 336: Let's not use `varient.x` -> move to a named tuple such as `.map(|(key, value)|`

sasial-dev (changes requested):

zap/src/config.rs
jackdotink: it actually shouldn't be zero, I'll fix it

zap/src/parser/convert.rs
```diff
@@ -146,17 +101,103 @@ impl<'src> SyntaxConfig<'src> {
...
+			let event_id_size = NumTy::from_f64(1.0, self.evdecls.len() as f64).size();
+
+			// We subtract two for the `inst` array.
+			let max_unreliable_size = 900 - event_id_size - 2;
+
+			if min > max_unreliable_size {
+				self.report(Report::AnalyzeOversizeUnreliable {
+					ev_span: evdecl.span(),
+					ty_span: evdecl.data.span(),
+					max_size: 900 - event_id_size,
```
sasial-dev, line 190:
```suggestion
					max_size: max_unreliable_size,
```

zap/src/parser/convert.rs
```diff
@@ -146,17 +101,103 @@ impl<'src> SyntaxConfig<'src> {
...
+					size: min,
+				});
+			} else if !max.is_some_and(|max| max < max_unreliable_size) {
+				self.report(Report::AnalyzePotentiallyOversizeUnreliable {
+					ev_span: evdecl.span(),
+					ty_span: evdecl.data.span(),
+					max_size: 900 - event_id_size,
```
sasial-dev, line 197:
```suggestion
					max_size: max_unreliable_size,
```

sasial-dev (changes requested):
More things - then we'll be ready to merge

zap/src/parser/convert.rs
sasial-dev: This is a terrible idea please don't

sasial-dev (approved):
It's time.

PR #52, opened 2024-01-04
Compress CFrames using axis-angle representation

Tazmondo:
Effectively halves the size of CFrames from 12 f32s down to 6 f32s.

Does not deal with non-orthogonal CFrames but this is not really a big issue as discussed in the discord channel.

EDIT: For future readers, I'll explain the reasoning behind this. 

- Non orthogonal CFrames can only really be made using the specific CFrame.new constructor where you specify every element of the matrix, or with CFrame.fromMatrix. Most users will never use these constructors as they are only used when doing complex math. They can also come about from accumulated error, but correcting that is probably for the best anyway.
- Furthermore, the use of non orthogonal CFrames are extremely limited - apparently being used mostly for skewing cameras. These use cases are rarely, if ever, going to require networking of the skewed CFrame.
- And finally, as documented here, https://github.com/Data-Oriented-House/Squash/pull/28, for the squash library, Roblox itself seems to orthonormalize CFrames before being passed over the network anyway. So, Zap isn't missing any features from Roblox's implementation anyway.

Please review harshly, I am still not super fluent with Rust so I don't know the common patterns and what good, clean, Rust code looks like.


zap/src/irgen/des.rs
```diff
@@ -281,16 +281,41 @@ impl Des {
...
+									vec!["axisangle".into(), "angle".into()],
+								)),
+								Box::new("pos".into()),
+							)),
+						)),
+						Box::new(Expr::Call(
+							Box::new(Var::from("CFrame").nindex("new")),
+							None,
+							vec!["pos".into()],
+						)),
```
sasial-dev, line 318: Let's move this over to be an `if`/`else` expression

sasial-dev (changes requested):
As discussed on Discord:

zap/src/irgen/des.rs
```diff
@@ -281,18 +281,44 @@ impl Des {
...
-				self.push_local("vY", Some(self.readvector3()));
-				self.push_local("vZ", Some(self.readvector3()));
+				self.push_local("axisangle", Some(self.readvector3()));
+				self.push_local(
+					"angle",
+					Some(Expr::Var(Box::new(Var::NameIndex(
+						Box::new("axisangle".into()),
+						"Magnitude".into(),
+					)))),
+				);
```
sasial-dev, line 291:
```suggestion
				self.push_local(
					"angle",
					Some(Expr::Var(Box::new(Var::from("axisangle").nindex("Magnitude")))),
				);
```

zap/src/output/luau/mod.rs
```diff
@@ -31,6 +31,22 @@ pub trait Output {
...
 			}
+			Stmt::LocalTuple(var, expr) => {
+				let mut items = String::new();
+
+				for (i, exp) in var.iter().enumerate() {
+					if i != 0 {
+						items.push_str(", ");
+					}
+					items.push_str(exp)
+				}
```
sasial-dev, line 42:
```suggestion
				let items = var.join(", ");
```

sasial-dev (changes requested):
Some initial comments

zap/src/irgen/des.rs
Tazmondo: Yeah that's a lot cleaner

sasial-dev (approved):
LGTM!

Tazmondo:
adding docs don't merge yet

jackdotink (approved):

PR #53, opened 2024-01-04
Disable RobloxLSP diagnostics

xSwezan:
I am simply proposing this change so that RobloxLSP users dont have a bunch of warnings in the outputs


sasial-dev:
Thanks for your PR!
As Roblox LSP is deprecated, I've decided to close this PR. I'd recommend using Luau LSP instead as it is maintained and uses luau instead of a seperate implementation under the hood.

PR #54, opened 2024-01-05
Add Remote Functions and Optional Data Field

### LINKED ISSUE #48: [FEAT] Make data field in events optional
## Describe your feature
Making `data` optional would enable sending remotes without any data.

## Alternatives
Adding a `void` type. This adds another type that isn't necessary and it's more difficult to maintain because you need to make sure it's being used correctly.

### LINKED ISSUE #49: [FEAT] Support RemoteFunctions
## Describe your feature
Events should be able to return a value

## Alternatives
Creating two events (one for each way)

## Implementation Details
Careful decision will have to be made as to if we use RemoteEvents under-the-hood and then create a mock remotefunction-like interface in zap (so we don't yield the thread!).

sasial-dev:
Closes #48 
Closes #49

### DIFF
--- a/zap/src/output/luau/client.rs
+++ b/zap/src/output/luau/client.rs
@@ -413,7 +413,10 @@ impl<'src> ClientOutput<'src> {
 	}
 
 	fn push_callback_lists(&mut self) {
-		self.push_line(&format!("local events = table.create({})", self.config.evdecls.len()));
+		self.push_line(&format!(
+			"local events = table.create({})",
+			self.config.evdecls.len() + self.config.fndecls.len()
+		));
 
 		if !self.config.fndecls.is_empty() {
 			self.push_line("local function_call_id = 0")
--- a/zap/src/output/luau/client.rs
+++ b/zap/src/output/luau/client.rs
@@ -135,7 +135,7 @@ impl<'src> ClientOutput<'src> {
 		));
 	}
 
-	fn push_reliable_callback(&mut self, first: bool, ev: &EvDecl, id: usize) {
+	fn push_reliable_callback(&mut self, first: bool, ev: &EvDecl) {
 		self.push_indent();
 
 		if first {
@@ -147,7 +147,7 @@ impl<'src> ClientOutput<'src> {
 		// push_line is not used here as indent was pushed above
 		// and we don't want to push it twice, especially after
 		// the if/elseif
-		self.push(&format!("id == {id} then"));
+		self.push(&format!("id == {} then", ev.id));
 		self.push("\n");
 
 		self.indent();
@@ -159,21 +159,21 @@ impl<'src> ClientOutput<'src> {
 		}
 
 		if ev.call == EvCall::SingleSync || ev.call == EvCall::SingleAsync {
-			self.push_line(&format!("if events[{id}] then"));
+			self.push_line(&format!("if events[{}] then", ev.id));
 		} else {
-			self.push_line(&format!("if events[{id}][1] then"));
+			self.push_line(&format!("if events[{}][1] then", ev.id));
 		}
 
 		self.indent();
 
 		if ev.call == EvCall::ManySync || ev.call == EvCall::ManyAsync {
-			self.push_line(&format!("for _, cb in events[{id}] do"));
+			self.push_line(&format!("for _, cb in events[{}] do", ev.id));
 			self.indent();
 		}
 
 		match ev.call {
-			EvCall::SingleSync => self.push_line(&format!("events[{id}](value)")),
-			EvCall::SingleAsync => self.push_line(&format!("task.spawn(events[{id}], value)")),
+			EvCall::SingleSync => self.push_line(&format!("events[{}](value)", ev.id)),
+			EvCall::SingleAsync => self.push_line(&format!("task.spawn(events[{}], value)", ev.id)),
 			EvCall::ManySync => self.push_line("cb(value)"),
 			EvCall::ManyAsync => self.push_line("task.spawn(cb, value)"),
 		}
@@ -188,18 +188,18 @@ impl<'src> ClientOutput<'src> {
 		self.indent();
 
 		if ev.data.is_some() {
-			self.push_line(&format!("table.insert(event_queue[{id}], value)"));
-			self.push_line(&format!("if #event_queue[{id}] > 64 then"));
+			self.push_line(&format!("table.insert(event_queue[{}], value)", ev.id));
+			self.push_line(&format!("if #event_queue[{}] > 64 then", ev.id));
 		} else {
-			self.push_line(&format!("event_queue[{id}] += 1"));
-			self.push_line(&format!("if event_queue[{id}] > 16 then"));
+			self.push_line(&format!("event_queue[{}] += 1", ev.id));
+			self.push_line(&format!("if event_queue[{}] > 16 then", ev.id));
 		}
 
 		self.indent();
 
 		self.push_line(&format!(
-			"warn(`[ZAP] {{#event_queue[{id}]}} events in queue for {}. Did you forget to attach a listener?`)",
-			ev.name
+			"warn(`[ZAP] {{#event_queue[{}]}} events in queue for {}. Did you forget to attach a listener?`)",
+			ev.id, ev.name
 		));
 
 		self.dedent();
@@ -211,7 +211,7 @@ impl<'src> ClientOutput<'src> {
 		self.dedent();
 	}
 
-	fn push_fn_callback(&mut self, first: bool, ev: &FnDecl, id: usize) {
+	fn push_fn_callback(&mut self, first: bool, fndecl: &FnDecl) {
 		self.push_indent();
 
 		if first {
@@ -223,7 +223,7 @@ impl<'src> ClientOutput<'src> {
 		// push_line is not used here as indent was pushed above
 		// and we don't want to push it twice, especially after
 		// the if/elseif
-		self.push(&format!("id == {id} then"));
+		self.push(&format!("id == {} then", fndecl.id));
 		self.push("\n");
 
 		self.indent();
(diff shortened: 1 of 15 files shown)


zap/src/output/luau/client.rs
```diff
@@ -184,9 +187,14 @@ impl<'src> ClientOutput<'src> {
...
 		self.indent();
 
-		self.push_line(&format!("table.insert(event_queue[{id}], value)"));
+		if ev.data.is_some() {
+			self.push_line(&format!("table.insert(event_queue[{id}], value)"));
+			self.push_line(&format!("if #event_queue[{id}] > 64 then"));
+		} else {
+			self.push_line(&format!("event_queue[{id}] += 1"));
+			self.push_line(&format!("if event_queue[{id}] > 16 then"));
+		}
```
jackdotink, line 196: We're using 64 for events with data, and 16 for events without data?

zap/src/output/luau/client.rs
```diff
@@ -300,9 +357,14 @@ impl<'src> ClientOutput<'src> {
...
 		self.indent();
 
-		self.push_line(&format!("table.insert(event_queue[{id}], value)"));
+		if ev.data.is_some() {
+			self.push_line(&format!("table.insert(event_queue[{id}], value)"));
+			self.push_line(&format!("if #event_queue[{id}] > 64 then"));
+		} else {
+			self.push_line(&format!("event_queue[{id}] += 1"));
+			self.push_line(&format!("if event_queue[{id}] > 64 then"));
+		}
```
jackdotink, line 366: same as above

zap/src/output/luau/client.rs
```diff
@@ -366,27 +435,39 @@ impl<'src> ClientOutput<'src> {
...
+			if ev_decl.data.is_some() {
+				self.push_line(&format!("event_queue[{id}] = {{}}"));
+			} else {
+				self.push_line(&format!("event_queue[{id}] = 0"));
+			}
+		}
+
+		for (i, ..) in self.config.fndecls.iter().enumerate() {
+			let id = i + 1;
 			self.push_line(&format!("event_queue[{id}] = {{}}"));
```
jackdotink, line 447: we should replace `{}` with `table.create(255)` to avoid resizing costs

zap/src/output/luau/client.rs
```diff
@@ -366,27 +435,39 @@ impl<'src> ClientOutput<'src> {
...
 
+			if ev_decl.data.is_some() {
+				self.push_line(&format!("event_queue[{id}] = {{}}"));
+			} else {
+				self.push_line(&format!("event_queue[{id}] = 0"));
+			}
+		}
+
+		for (i, ..) in self.config.fndecls.iter().enumerate() {
+			let id = i + 1;
```
jackdotink, line 446: `i` cannot be used as a means of getting event id because it will intersect with evdecls

zap/src/output/luau/client.rs
```diff
@@ -517,6 +636,85 @@ impl<'src> ClientOutput<'src> {
...
+
+			self.push_line("function_call_id += 1");
+
+			self.push_line("function_call_id %= 256");
+
+			self.push_line(&format!("if event_queue[{id}][function_call_id] then"));
+			self.indent();
+
+			self.push_line("function_call_id -= 1");
+			self.push_line("error(\"Zap has more than 256 events awaiting a response, and therefore this packet has been dropped.\")");
```
jackdotink, line 670: change `events` to `calls`

zap/src/output/luau/client.rs
```diff
@@ -517,6 +636,85 @@ impl<'src> ClientOutput<'src> {
...
+			self.push_line("buffer.writeu8(outgoing_buff, outgoing_apos, function_call_id)");
+
+			if let Some(data) = &fndecl.args {
+				self.push_stmts(&ser::gen(data, value, self.config.write_checks));
+			}
+
+			match self.config.yield_type {
+				YieldType::Yield => {
+					self.push_line(&format!("event_queue[{id}][function_call_id] = coroutine.running()"));
+					self.push_line("local value = coroutine.yield()");
```
jackdotink, line 685: we want to `return coroutine.yield()`, not just call it

zap/src/output/luau/client.rs
```diff
@@ -550,11 +749,12 @@ impl<'src> ClientOutput<'src> {
...
-			.config
-			.evdecls
-			.iter()
-			.any(|ev| ev.evty == EvType::Reliable && ev.from == EvSource::Server)
+		if !self.config.fndecls.is_empty()
+			|| self
+				.config
+				.evdecls
+				.iter()
+				.any(|ev| ev.evty == EvType::Reliable && ev.from == EvSource::Server)
```
jackdotink, line 757: iirc not has a lower precedence than or, so we should wrap the or in parens and then use not. I might be wrong on that though

zap/src/output/luau/server.rs
```diff
@@ -201,6 +259,13 @@ impl<'a> ServerOutput<'a> {
 			first = false;
 		}
 
+		for (i, fndecl) in self.config.fndecls.iter().enumerate() {
+			let id = i + 1;
```
jackdotink, line 263: once again, we need to find a new way to ids

zap/src/output/luau/server.rs
```diff
@@ -608,11 +752,12 @@ impl<'a> ServerOutput<'a> {
...
-			.config
-			.evdecls
-			.iter()
-			.any(|ev| ev.evty == EvType::Reliable && ev.from == EvSource::Client)
+		if !self.config.fndecls.is_empty()
+			|| self
+				.config
+				.evdecls
+				.iter()
+				.any(|ev| ev.evty == EvType::Reliable && ev.from == EvSource::Client)
```
jackdotink, line 760: same as `client.rs`

zap/src/parser/convert.rs
```diff
@@ -176,28 +210,30 @@ impl<'src> Converter<'src> {
...
-			} else if !max.is_some_and(|max| max < max_unreliable_size) {
-				self.report(Report::AnalyzePotentiallyOversizeUnreliable {
-					ev_span: evdecl.span(),
-					ty_span: evdecl.data.span(),
-					max_size: max_unreliable_size,
-				});
+		let data = evdecl.data.as_ref().map(|ty| self.ty(ty));
+
+		if let Some(data) = &data {
+			if let EvType::Unreliable = evty {
```
jackdotink, line 216:
```suggestion
		if data.is_some() && evty == EvType::Unreliable {
			let data = data.unwrap();
```

jackdotink (changes requested):

zap/src/output/luau/client.rs
sasial-dev: but won't that use extra memory? or is that just not how it works

zap/src/output/luau/client.rs
sasial-dev: This was a deliberate choice I mentioned in the thread. The existing choice of delivering 64 events w/o data is fine. But I feel like running 64 callbacks w/o data doesn't feel right.

My logic is that if there's no data inside the event - then the event shouldn't fire as often, and if we have the run the same callback without args 64 times then something's gone wrong...

zap/src/output/luau/client.rs
sasial-dev:
```suggestion
			self.push_line("error(\"Zap has more than 256 calls awaiting a response, and therefore this packet has been dropped\")");
```

zap/src/output/luau/client.rs
sasial-dev: Fixed the bug that Promise/Future didn't assign `value`. Below the match it returns `value`.

zap/src/output/luau/client.rs
sasial-dev: Should we assign in ID in `convert.rs` instead?

zap/src/output/luau/client.rs
jackdotink: using a constant amount of memory (especially when it's not that much) is often better than dynamically resizing the array. Also with this, unless we statically size the entire array it's almost certain that when we index it we hit the hashmap part, not the array part.

zap/src/output/luau/client.rs
jackdotink: sgtm

zap/src/parser/convert.rs
sasial-dev: I personally think it's less clear this way - what's your intent here?

jackdotink: less nesting is usually better, and if we're checking two conditions we shouldn't do it with two if statements

zap/src/output/luau/client.rs
sasial-dev: That's what I want!

zap/src/parser/reports.rs
```diff
@@ -298,6 +310,12 @@ impl<'src> Report<'src> {
 				"this is an unbounded recursive type".to_string(),
 				"unbounded recursive types cause infinite loops".to_string(),
 			]),
+			Self::AnalyzeMissingOptValue {
+				expected,
+				required_when: required_by,
+			} => Some(vec![format!(
+				"the {expected} option should not be empty if {required_by}"
+			)]),
```
jackdotink, line 318: keep the name as `required_when`, no need to change it to `required_by`

jackdotink (approved):

zap/src/parser/reports.rs
sasial-dev: Oh that was when I renamed the var

sasial-dev:
```suggestion
			Self::AnalyzeMissingOptValue {
				expected,
				required_when,
			} => Some(vec![format!(
				"the {expected} option should not be empty if {required_when}"
			)]),
```

PR #73, opened 2024-03-01
enable zap development on nixos using a nix shell

AbleTheAbove:
This is useful to allow certain rust developers, namely myself, to contribute in a way that does not use globally installed rust.


AbleTheAbove:
git is not the good. Copy paste better 🗣️
I'll reopen after nuking my branch and rebaseing.

Issue #71, opened 2024-02-12
[BUG] TypeScript type generation should use `declare` keyword


osyrisrblx:
## Describe the bug
Given an input like
```
opt typescript = true

event MyEvent = {
	from: Server,
	type: Reliable,
	call: ManyAsync,
	data: u8,
}
```

Zap will generate
```ts
// Server generated by Zap v0.6.1 (https://github.com/red-blox/zap)
export const MyEvent: {
	Fire: (Player: Player, Value: number) => void
	FireAll: (Value: number) => void
	FireExcept: (Except: Player, Value: number) => void
	FireList: (List: Player[], Value: number) => void
};
```

This is missing the `declare` keyword. Not technically a requirement, but good practice.
[TypeScript docs](https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html)

## Reproduction
[playground link](https://zap.redblox.dev/playground.html?code=b3B0IHR5cGVzY3JpcHQgPSB0cnVlCgpldmVudCBNeUV2ZW50ID0gewoJZnJvbTogU2VydmVyLAoJdHlwZTogUmVsaWFibGUsCgljYWxsOiBNYW55QXN5bmMsCglkYXRhOiB1OCwKfQ==)

## Expected behavior

```ts
// Server generated by Zap v0.6.1 (https://github.com/red-blox/zap)
export declare const MyEvent: {
	Fire: (Player: Player, Value: number) => void
	FireAll: (Value: number) => void
	FireExcept: (Except: Player, Value: number) => void
	FireList: (List: Player[], Value: number) => void
};
```

PR #76, opened 2024-03-19
fix(typescript): use export declare const instead

fisherjacobc:
I believe that it is better practice to use `export declare const` instead for the declaration files, since there isn't a value assigned to it.

I saw it in #71 and thought I could help out since it's a simple fix


sasial-dev:
Thank you so much for helping out!
So here's the deal - we're currently undergoing a rewrite of zap (see `rewrite` branch). And, as this isn't a bug - I don't believe it's worth backporting this fix to the `0.6.x` branch.

sasial-dev:
I was going to rebase your PR to fit into `0.6.4`, but I do not have permission to do so. You're more than welcome to rebase it yourself  after fetching down the branch (`git rebase -i 0.6.x`, and then drop all commits but the one you made) and then I'll merge it in/

fisherjacobc:
If you want to go ahead and do it, that might be easier so I don't mess something up. You should have write permissions now.

sasial-dev:
It still doesn't seem to give me permisison. Might be eaisier to just take your diff from https://github.com/red-blox/zap/pull/76/commits/a9856a33286ad1d02f6d5743eb7bc5a7ac5ea438 and send another PR based on the `0.6.x` branch.

fisherjacobc:
Refer to #103

PR #79, opened 2024-03-31
Optional `types_output` option

### LINKED ISSUE #77: [FEAT] Add option for exporting zap types into separate file.
## Describe your feature
Currently types you write inside your zap file get exported in both the server and client files, this is usually fine. However, I'm pretty nitpicky about separation, and I am not a fan of having my shared files access client files just to access types.

## Implementation Details
`opt types_output = "src/shared/types/zap.types.luau`
This would create a new file in the location specified, and then your server & client would gather the types necessary from there.

passhley:
Adds a new option `types_output` to export types written in your Zap file out to a shared file.

Closes #77


zap/src/parser/convert.rs
```diff
@@ -207,6 +209,25 @@ impl<'src> Converter<'src> {
...
+				self.report(Report::AnalyzeInvalidOptValue {
+					span: opt.value.span(),
+					expected: "string",
+				});
+			}
+		}
+
+		(value, span)
+	}
+
```
jackdotink, line 230: I'd rather not add `optional_str_opt` and instead have a `str_opt` with an empty default (`""`). Later we can check `types_output.is_empty()` to determine if we should output.

passhley: Updated `types_output` to use `str_opt` instead.

sasial-dev:
I had a go trying to rebase this but I wasn't sucessful. The same code should almost work - but could you please base your PR off the `0.6.x` branch? I apologise for the lack of clarity, and I'll update it on the README now!

PR #90, opened 2024-06-11
[REWRITE] More errors

nezuo:


zapc/src/meta/report.rs
```diff
@@ -151,6 +159,30 @@ impl Report {
...
+			Self::DuplicateDecl {
+				decl_kind,
+				name,
+				span,
+				first_decl_span,
+			} => build(kind, span)
+				.with_message(format!(
+					"the {} {} is defined multiple times in the same scope",
+					decl_kind,
+					ticks(&name)
```
jackdotink, line 172: change to `ticks(&name).fg(ERROR)`

zapc/src/meta/report.rs
```diff
@@ -151,6 +159,30 @@ impl Report {
...
+				))
+				.with_labels([
+					label(first_decl_span)
+						.with_message(format!(
+							"first definition of the {} {} here",
```
jackdotink, line 177: `declkind name is first defined here`

zapc/src/meta/report.rs
```diff
@@ -151,6 +159,30 @@ impl Report {
...
+							decl_kind,
+							ticks(&name).fg(INFO)
+						))
+						.with_color(INFO),
+					label(span)
+						.with_color(ERROR)
+						.with_message(format!("{} redefined here", ticks(&name).fg(ERROR))),
```
jackdotink, line 184: declkind name later redefined here

zapc/src/hir_builder/decl.rs
```diff
@@ -1,45 +1,67 @@
...
-		}
+		});
+
+		self.report_duplicates("remote", &decls, |d| {
+			if let AstDecl::Remote { name, .. } = d {
+				Some((name, name.span()))
+			} else {
+				None
+			}
+		});
```
jackdotink, line 64: In `hir_builder/scope.rs` I have some code for reporting errors should duplicate types or remotes attempt to be added. I do agree that reporting all duplicates here is far better, but in those other cases it needs to be ignored.

zapc/src/hir_builder/decl.rs
```diff
@@ -59,148 +60,238 @@ impl<'a> HirBuilder<'a> {
...
-					// todo: report error
 				}
 
+				self.report(Report::ExpectedValueWrongType {
+					value_span: words.first().unwrap().span().merge(words.last().unwrap().span()),
+					expected_values: "\"client\" or \"server\"".to_string(),
+					found_type: "path".to_string(),
+					help: None,
+				});
+
```
jackhexed: we will almost certainly not reuse `ExpectedValueWrongType`, so I think it would be a better idea to create a more specific report variant for this case.

zapc/src/hir_builder/ty.rs
```diff
@@ -91,9 +114,16 @@ impl<'a> HirBuilder<'a> {
 		})
 	}
 
-	fn generics_one_range(&mut self, generics: &[AstGeneric]) -> Option<AstRange> {
+	fn generics_one_range(&mut self, segment: &AstWord, generics: &[AstGeneric]) -> Option<AstRange> {
 		if generics.len() > 1 {
-			// todo: report extra generics
+			self.report(Report::IncorrectGenericCount {
+				type_span: segment.span(),
```
jackhexed: this wouldn't be the span of the entire type
imagine the type `buffer<a, b>` - this would only be the span for `buffer` when you really want it to be the span for the whole type

zapc/src/meta/report.rs
```diff
@@ -151,6 +248,173 @@ impl Report {
...
+								"remove this generic argument"
+							} else {
+								"remove these generic arguments"
+							}),
+					);
+				}
+
+				build(kind, type_span)
+					.with_message(format!(
+						"type {} takes {}{expected_count} generic {} but {count} generic {} supplied",
```
jackhexed, line 404: "generic" should be "generics" when `count > 1`

jackhexed (changes requested):

zapc/src/meta/report.rs
nezuo: This is already accounted for by the following word: `type x takes 5 generic arguments`

nezuo:
CI was failing because of Rustfmt. It wasn't a file I touched but I formatted it anyways.

sasial-dev:
Thank you for your contributions so far!
The state of the rewrite is tenuous, so I'll close this PR. Feel free to keep the branch in case things change.

PR #93, opened 2024-06-23
Fix nested maps

### LINKED ISSUE #89: [BUG] Zap incorrectly shadows variables which causes deserialization to fail
## Describe the bug
Zap de-serialization will shadow variables and try to use the value of the variable before it was shadowed

## Reproduction
https://zap.redblox.dev/playground.html?code=ZXZlbnQgZXhhbXBsZSA9IHsNCiAgICBmcm9tOiBTZXJ2ZXIsDQogICAgdHlwZTogUmVsaWFibGUsDQogICAgY2FsbDogU2luZ2xlU3luYywNCiAgICBkYXRhOiBtYXAgeyBbdThdOiBtYXAgeyBbdThdOiBib29sZWFuIH0gfQ0KfQ0K

If you look at the client you will see:
```lua
local key
local val
key = buffer.readu8(incoming_buff, read(1))
val = {}
for _ = 1, buffer.readu16(incoming_buff, read(2)) do
	local key
	local val
	key = buffer.readu8(incoming_buff, read(1))
	val = buffer.readu8(incoming_buff, read(1)) == 1 -- shadows the `val` table
	val[key] = val -- errors because `val` is now a boolean instead of the table
end
```

Ketasaja:
Removes shadowing from generated code, allowing for events like:
```
event test = {
	from: Client,
	type: Reliable,
	call: ManyAsync,
	data: map { [
		map { [string]: boolean }
	]:
		map {
			[boolean]:
			string[][][]
		}
	}[]
}
```

Ideally there'd be an integer suffix instead of `_v_v_v_v_v_v_v`, this is a quick fix to hold over until the rewrite.

### DIFF
--- a/zap/src/irgen/des.rs
+++ b/zap/src/irgen/des.rs
@@ -155,13 +155,17 @@ impl Des {
 					to: self.readu16(),
 				});
 
-				self.push_local("key", None);
-				self.push_local("val", None);
+				let escaped = into.display_escaped();
+				let suffix = &escaped[escaped.find("_").unwrap()..];
+				let key_name = "key".to_string() + suffix;
+				self.push_local(key_name.clone().leak(), None);
+				let val_name = "val".to_string() + suffix;
+				self.push_local(val_name.clone().leak(), None);
 
-				self.push_ty(key, "key".into());
-				self.push_ty(val, "val".into());
+				self.push_ty(key, Var::Name(key_name.clone()));
+				self.push_ty(val, Var::Name(val_name.clone()));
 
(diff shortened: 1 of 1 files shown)


sasial-dev:
Oh. My. God - you are awesome, thank you for doing this!

sasial-dev (approved):
I don't like the idea of using `.leak()`, but I think it's okay considering we also leak in array des (L139). I did mess around with trying to get rid of it but I had no success.
Otherwise, I really like the way you did the solution - it's pretty clever!

PR #95, opened 2024-06-24
Generate TypeScript arrays instead of huge tuples

### LINKED ISSUE #69: [BUG] TypeScript type generation should only use tuples for small arrays
## Describe the bug

Given an input like this
```
opt typescript = true

event MyEvent = {
	from: Server,
	type: Reliable,
	call: ManyAsync,
	data: u8[5],
}
```
Zap will generate TypeScript types like:
```ts
Fire: (Player: Player, Value: [number, number, number, number, number]) => void
```

However.. if you use something like `data: u8[128][128][128]`, this very quickly gets out of hand with 2 million `number` types.

Instead, if an array type exceeds a certain size, Zap should just emit `number[]`.

## Reproduction
[playground link](https://zap.redblox.dev/playground.html?code=b3B0IHR5cGVzY3JpcHQgPSB0cnVlCgpldmVudCBNeUV2ZW50ID0gewoJZnJvbTogU2VydmVyLAoJdHlwZTogUmVsaWFibGUsCgljYWxsOiBNYW55QXN5bmMsCglkYXRhOiB1OFsxMjhdWzEyOF1bMTI4XSwKfQo=)

## Expected behavior
Zap should emit TypeScript types which don't make the IDE explode.

Ketasaja:
Closes #69.

This doesn't have special behavior for nested types like `u8[10][10][10][10]` to keep this uncontroversial.

Constant `const TUPLE_BECOMES_ARRAY_THRESHOLD: i8 = 11;` specifies the threshold. Alternatively there could be a configuration option for this.

### DIFF
--- a/zap/src/output/typescript/mod.rs
+++ b/zap/src/output/typescript/mod.rs
@@ -4,7 +4,7 @@ pub mod client;
 pub mod server;
 
 pub trait ConfigProvider {
-    fn get_config(&self) -> &Config;
+	fn get_config(&self) -> &Config;
 }
 
 pub trait Output: ConfigProvider {
--- a/zap/src/output/typescript/mod.rs
+++ b/zap/src/output/typescript/mod.rs
@@ -3,6 +3,8 @@ use crate::config::{Config, Enum, Ty};
 pub mod client;
 pub mod server;
 
+const TUPLE_BECOMES_ARRAY_THRESHOLD: i8 = 11;
+
 pub trait Output {
 	fn push(&mut self, s: &str);
 	fn indent(&mut self);
@@ -24,17 +26,22 @@ pub trait Output {
 			Ty::Arr(ty, range) => match (range.min(), range.max()) {
 				(Some(min), Some(max)) => {
 					if let Some(exact) = range.exact() {
-						self.push("[");
+						if exact >= TUPLE_BECOMES_ARRAY_THRESHOLD.into() {
+							self.push_ty(ty);
+							self.push("[]");
+						} else {
+							self.push("[");
 
-						for i in 0..exact as usize {
-							if i != 0 {
-								self.push(", ");
+							for i in 0..exact as usize {
+								if i != 0 {
+									self.push(", ");
+								}
+
+								self.push_ty(ty);
 							}
 
-							self.push_ty(ty);
+							self.push("]");
 						}
-
-						self.push("]");
 					} else {
 						if min as usize != 0 {
 							self.push("[");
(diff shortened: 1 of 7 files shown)


zap/src/output/typescript/mod.rs
```diff
@@ -3,6 +3,8 @@ use crate::config::{Config, Enum, Ty};
 pub mod client;
 pub mod server;
 
+const TUPLE_BECOMES_ARRAY_THRESHOLD: i8 = 11;
```
sasial-dev: Let's turn this into a configuration option.

sasial-dev (changes requested):

zap/src/output/typescript/mod.rs
Ketasaja: What should it be called? `typescript_tuples_become_arrays_threshold`, or is that too verbose? Luau might offer tuple types someday and I don't want to confuse anyone.
luau tuples discord

sasial-dev: I think we'll need to document it either way (see `docs/options.md`).
Maybe `typescript_array_threshold`?

zap/src/output/typescript/mod.rs
```diff
@@ -185,18 +190,18 @@ pub trait Output {
 		}
 	}
 
-	fn push_arg_ty(&mut self, ty: &Ty) {
+	fn push_arg_ty(&mut self, ty: &Ty, config: &Config) {
```
sasial-dev: Let's not make `config` an argument of `push_ty`/`push_arg_ty`. ~Instead add it to the trait itself~ - I realised you can't do this in rust. I'll try to think about alternative options.

zap/src/parser/convert.rs
```diff
@@ -88,6 +88,7 @@ impl<'src> Converter<'src> {
 		let (write_checks, ..) = self.boolean_opt("write_checks", true, &config.opts);
 		let (typescript, ..) = self.boolean_opt("typescript", false, &config.opts);
 		let (manual_event_loop, ..) = self.boolean_opt("manual_event_loop", false, &config.opts);
+		let (typescript_max_tuple_length, ..) = self.num_opt("typescript_max_tuple_length", 11.0, &config.opts);
```
sasial-dev: Shouldn't this be `10.0` as that's what you specified the deafult was on the docs.

sasial-dev (changes requested):

sasial-dev:
After doing those commits please also fix the merge conflicts.

sasial-dev (approved):
LGTM!

PR #97, opened 2024-06-24
Freeze module returns

gaymeowing:
Makes the return table of the generated modules frozen, including the nested tables for events

### DIFF
--- a/zap/src/output/luau/client.rs
+++ b/zap/src/output/luau/client.rs
@@ -576,7 +576,7 @@ impl<'src> ClientOutput<'src> {
 			.iter()
 			.filter(|ev_decl| ev_decl.from == EvSource::Client)
 		{
-			self.push_line(&format!("{name} = {{", name = ev.name));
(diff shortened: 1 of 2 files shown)


jackhexed (approved):
lgtm

sasial-dev (changes requested):
Small nit with regards to code style, you use `table.freeze({` sometimes and sometimes you use `table.freeze {`. Can you convert all of them to be `table.freeze({`?

gaymeowing:
> Small nit with regards to code style, you use `table.freeze({` sometimes and sometimes you use `table.freeze {`. Can you convert all of them to be `table.freeze({`?

Only did table.freeze({ for the studio return, as table.freeze {} :: T in the new solver currently produces a type error. Will do it if you still think its necessary, I just think having parentheses would unnecessarily clutter up how the return looks.

sasial-dev (approved):
LGTM!

PR #98, opened 2024-06-24
Add `FireSet` method

sasial-dev:


sasial-dev:
@nezuo can we assume that sets always have the value as `true` (so I can just do `for i in set`) or do we have to evaluate the key's value at runtime?

nezuo:
Yes, it should assume that the value is always `true`. It will also be type-checked at compile time.

jackhexed (approved):
lgtm

PR #100, opened 2024-06-25
Add DateTime types

gaymeowing:
Adds in DateTime and DateTimeMillis as types, with DateTime sending the UnixTimestamp property of the DateTime object, and DateTimeMillis sending the UnixTimestampMillis property of the DateTime object.

### DIFF
--- a/docs/config/types.md
+++ b/docs/config/types.md
@@ -215,11 +215,16 @@ local CFrameSpecialCases = {
 }
 ```
 
+## DateTimes
+
+Zap supports sending DateTimes. There are two types of DateTime you may send - a regular `DateTime`, and `DateTimeMillis`.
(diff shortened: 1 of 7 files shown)


sasial-dev (commented):
I'm pretty happy with the code. I just would like you to confirm - what's the motivation between adding two DateTime types? I definitely think we'll keep it, just want to make sure we're on the same wavelength!

gaymeowing:
> I'm pretty happy with the code. I just would like you to confirm - what's the motivation between adding two DateTime types? I definitely think we'll keep it, just want to make sure we're on the same wavelength!

Mainly because sometimes using milliseconds is desired over seconds for cases where the extra accuracy is needed

sasial-dev (changes requested):
Please add the types to be matched in `parser/convert.rs` to at L416 (under `SyntaxTyKind::Ref`).

sasial-dev (approved):
LGTM!

PR #101, opened 2024-06-25
Add BrickColor type

gaymeowing:
Adds a BrickColor type to zap, as BrickColors can be serialized more efficiently than Color3s as they can be serialized as only a u16

### DIFF
--- a/zap/src/irgen/ser.rs
+++ b/zap/src/irgen/ser.rs
@@ -248,6 +248,12 @@ impl Ser {
 
 			Ty::BrickColor => self.push_writeu16(from.clone().nindex("Number").into()),
 
+			Ty::Vector3 => {
+				self.push_writef32(from.clone().nindex("X").into());
+				self.push_writef32(from.clone().nindex("Y").into());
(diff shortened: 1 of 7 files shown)


zap/src/irgen/ser.rs
```diff
@@ -246,11 +246,7 @@ impl Ser {
 				));
 			}
 
-			Ty::Vector3 => {
-				self.push_writef32(from.clone().nindex("X").into());
-				self.push_writef32(from.clone().nindex("Y").into());
-				self.push_writef32(from.clone().nindex("Z").into());
-			}
+			Ty::BrickColor => self.push_writeu16(from.clone().nindex("Number").into()),
```
sasial-dev: I think you unintentionally got rid of `Vector3`.

sasial-dev (changes requested):

sasial-dev (changes requested):
Please add the types to be matched in `parser/convert.rs` to at L416 (under `SyntaxTyKind::Ref`).

sasial-dev (approved):
LGTM!

Issue #78, opened 2024-03-26
[FEAT] Vector2 class support


Blueeyy:
## Describe your feature
Support for the Vector2 class; this is more of a personal need so I apologize in advance. I use Zap for voxel replication and I only need to save 2 values instead of 3.

jackdotink:
For transparency here zap will never support the Vector2 class, but will instead support Vector3's with two values. This is due to luau heap-allocating Vector2's and stack-allocating Vector3's. This means they have a dramatic performance difference.

PR #102, opened 2024-06-25
Add Vector2 type

gaymeowing:
Closes https://github.com/red-blox/zap/issues/78, adds in a Vector2 type to zap, same as the Vector3 type except that it doesn't serialize the Z axis.

### DIFF
--- a/docs/config/types.md
+++ b/docs/config/types.md
@@ -215,11 +215,15 @@ local CFrameSpecialCases = {
 }
 ```
 
+## Vectors
+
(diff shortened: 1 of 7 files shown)


sasial-dev (commented):
The code is pretty much good but I want to check with @jackdotink as I disagree and don't think we should force our users to use `Vector2`'s. Although we try and make our code as performant as possible, we shouldn't force any changes in the user's where possible. This should include `Vector2`/`Vector3` 'lockin'.

jackhexed:
Roblox Vector2s are not actually useful for much of anything. In the cases where the API requires you to use them, you aren't really networking data. In the case where you want a vector with two components, they aren't useful because they're heap allocated and that comes with a lot of baggage.

Instead we should be using the Luau `vector` datatype, and let the user decide how many components to send. I've made some revisions to the Luau vector library RFC, so that should hopefully happen soon.

Additionally, the way I want to design this would only really work in the rewrite.

sasial-dev (approved):
LGTM!

PR #103, opened 2024-06-25
fix(typescript): use export declare const instead

### LINKED ISSUE #71: [BUG] TypeScript type generation should use `declare` keyword

fisherjacobc:
Resolves #71 
Attempt 2 of #76

### DIFF
--- a/zap/src/output/typescript/client.rs
+++ b/zap/src/output/typescript/client.rs
@@ -68,7 +68,7 @@ impl<'src> ClientOutput<'src> {
 			let fire = self.config.casing.with("Fire", "fire", "fire");
 			let value = self.config.casing.with("Value", "value", "value");
 
-			self.push_line(&format!("export const {name}: {{", name = ev.name));
+			self.push_line(&format!("export declare const {name}: {{", name = ev.name));
 			self.indent();
 
 			self.push_indent();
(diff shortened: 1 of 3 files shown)


sasial-dev (approved):
Thank you!

PR #104, opened 2024-06-26
Remove benign shadowing

Ketasaja:
Suffixes are less incremental than with `display_escaped_suffix`, but there should be no shadowing left. I'm not aware of this being necessary to fix anything, unlike previous changes.


zap/src/irgen/des.rs
```diff
@@ -86,34 +86,36 @@ impl Des {
 				if let Some(len) = range.exact() {
 					self.push_assign(into, self.readstring(len.into()));
 				} else {
-					self.push_local("len", Some(self.readnumty(NumTy::U16)));
+					let len_name = format!("len_{}", self.buf.len());
```
sasial-dev: Why did you go with the string length instead of just counting up each time `len` is used for example? It could be a HashMap or something in `irgen`?

zap/src/irgen/mod.rs
```diff
@@ -1,11 +1,29 @@
 #![allow(dead_code)]
+use std::collections::HashMap;
 use std::{fmt::Display, vec};
 
 use crate::config::{NumTy, Range, Ty};
 
 pub mod des;
 pub mod ser;
 
+pub trait VarOccurrencesProvider {
```
sasial-dev: Why did you choose to use a seperate trait instead of making it a part of `Gen`?

zap/src/irgen/des.rs
```diff
@@ -1,10 +1,12 @@
 use crate::config::{Enum, NumTy, Struct, Ty};
+use std::collections::HashMap;
 
-use super::{Expr, Gen, Stmt, Var};
+use super::{Expr, Gen, Stmt, Var, VarOccurrencesProvider};
 
 struct Des {
 	checks: bool,
 	buf: Vec<Stmt>,
+	var_occurrences: HashMap<String, i32>,
```
sasial-dev: This `i32` should be a `usize`?

zap/src/irgen/des.rs
```diff
@@ -86,34 +88,36 @@ impl Des {
 				if let Some(len) = range.exact() {
 					self.push_assign(into, self.readstring(len.into()));
 				} else {
-					self.push_local("len", Some(self.readnumty(NumTy::U16)));
+					let len_name = format!("len_{}", self.add_occurrence("len".into()));
```
sasial-dev: Maybe `add_occurrence` should return a `Var` or an `Expr` from the name?

zap/src/irgen/mod.rs
```diff
@@ -1,11 +1,29 @@
...
+		match self.get_var_occurrences().get(&name) {
+			Some(occurrences) => {
+				let occurrences_inc = occurrences + 1;
+				self.get_var_occurrences().insert(name, occurrences_inc);
+				occurrences_inc
+			}
+			None => {
+				self.get_var_occurrences().insert(name, 1);
+				1
+			}
```
sasial-dev, line 22: If you follow the above question where it returns the value, if this `None` clause is reached, it should just return the var name.
IE: `len_1` -> `len`.

sasial-dev (changes requested):

zap/src/irgen/mod.rs
Ketasaja: That can create shadowing again, like with the header.
`self.push_line("local len = buffer.len(buff)");`

sasial-dev: oh okay! no worries then

zap/src/irgen/des.rs
```diff
@@ -30,7 +36,7 @@ impl Des {
 			Enum::Unit(enumerators) => {
 				let numty = NumTy::from_f64(0.0, enumerators.len() as f64 - 1.0);
 
-				self.push_local("enum_value", Some(self.readnumty(numty)));
+				self.push_local("enum_value".into(), Some(self.readnumty(numty)));
```
sasial-dev: Maybe we move all vars to use `add_occurence`, or make `push_local` use it under the hood? `enum_value` can shadow as shown in [this example](https://zap.redblox.dev/playground?code=ZXZlbnQgZXhhbXBsZSA9IHsKICAgIGZyb206IFNlcnZlciwKICAgIHR5cGU6IFJlbGlhYmxlLAogICAgY2FsbDogU2luZ2xlU3luYywKICAgIGRhdGE6IGVudW0gInRlc3QiIHsKICAgICAgICBBIHsKICAgICAgICAgICAgZm9vOiBlbnVtIHsgYSwgYiwgYyB9CiAgICAgICAgfQogICAgfQp9Cg==).

sasial-dev (changes requested):

zap/src/irgen/des.rs
Ketasaja: The problem with modifying `push_local` is stuff like this. `push_local` could check if the name is already suffixed and do nothing, or all variables could use `add_occurrence`.
```rs
let (len_name, len_expr) = self.add_occurrence("len");
self.push_local(
  len_name.clone(),
  Some(
	  Var::from("buffer")
		  .nindex(len_name.clone())
		  .call(vec![from_expr.clone()]),
  ),
);
```

Ketasaja:
I kind of dislike the complexity being added here, and I still haven't handled cases like `Stmt::LocalTuple` or `Stmt::Assign`.

sasial-dev:
Maybe let's revert that commit then? I agree it's not the greatest approch to hide it.

jackhexed (approved):
lgtm

sasial-dev (approved):
LGTM!

PR #107, opened 2024-06-26
Implement remote scope option

### LINKED ISSUE #106: [FEAT] Support multiple instances of zap within one place
## Describe your feature
I'm suggesting there is a new `opt` parameter to allow segmenting multiple zap net codes:

`opt remote_scope = "name"`

where this option will modify the remote emits from:

`ZAP_UNRELIABLE`
`ZAP_RELIABLE`

to:

`name_UNRELIABLE`
`name_RELIABLE`

or some other reasonable naming convention.

That way any conflicts between packages and place net code can be resolved.

## Alternatives
There is no known alternatives as of writing that would allow this functionality.

## Implementation Details
Implementation should change the emit of the remote name based on whether or not this option is set. The implementation is relatively simple as it involves just adding a condition to check if an option is set and changing the remote name in emits.

## Additional context
Right now, zap can only be used once within a given place file. While this is useful enough for many people, this implementation, especially as a package writer, this is not enough. Zap should be able to segment net code in packages away from other net code in other places.

ghost:
Changes the way zap remotes are created.

If the `remote_scope` option is set, the remote name will replace `ZAP` in remote names with the option set.

Closes #106. The motivation for this PR can be found in that issue.


sasial-dev (changes requested):
I'd rather `SCOPE_NAME` was hardcoded in on the rust side instead of using template strings. If you need/want to delete `client.luau` and `server.luau` to do so, that doesn't bother me.

zap/src/output/luau/server.luau
```diff
@@ -1,22 +1,4 @@
...
-end
-
-local unreliable = ReplicatedStorage:FindFirstChild("ZAP_UNRELIABLE")
-if unreliable == nil then
-	unreliable = Instance.new("UnreliableRemoteEvent")
-	unreliable.Name = "ZAP_UNRELIABLE"
-	unreliable.Parent = ReplicatedStorage
-end
 
 local player_map = {}
```
sasial-dev: Let's completely get rid of `server.luau` if you're getting rid of `client.luau`.

zap/src/output/luau/client.rs
```diff
@@ -839,7 +855,11 @@ impl<'src> ClientOutput<'src> {
 
 		self.push_studio();
 
-		self.push(include_str!("client.luau"));
+		self.push_check_server();
+
+		self.push_remotes();
+
+		self.push_line("local time = 0");
```
sasial-dev: This is... awkward! Let's move this whole block of push functions you added into one function call maybe?

zap/src/output/luau/server.rs
```diff
@@ -888,6 +918,12 @@ impl<'a> ServerOutput<'a> {
 
 		self.push_studio();
 
+		self.push_line("local Players = game:GetService(\"Players\")");
```
sasial-dev: This is... awkward! Let's move this whole block of push functions you added into one function call maybe?

zap/src/parser/convert.rs
```diff
@@ -92,6 +92,7 @@ impl<'src> Converter<'src> {
 
 		let (server_output, ..) = self.str_opt("server_output", "network/server.lua", &config.opts);
 		let (client_output, ..) = self.str_opt("client_output", "network/client.lua", &config.opts);
+		let (remote_scope, ..) = self.str_opt("remote_scope", "ZAP", &config.opts);
```
sasial-dev, line 95: Add this option (and if you could do me a favour and reconcille any missing ones) to `docs/.vitepress/components/Editor.vue` on L117.

sasial-dev (changes requested):

zap/src/output/luau/server.luau
ghost: Sounds good to me

zap/src/output/luau/client.rs
```diff
@@ -828,6 +829,23 @@ impl<'src> ClientOutput<'src> {
 		self.push_line("return returns");
 	}
 
+	pub fn push_remotes(&mut self) {
+		self.push_line(&format!("local reliable = ReplicatedStorage:WaitForChild(\"{}_RELIABLE\")", self.config.remote_scope));
+		self.push_line(&format!("local unreliable = ReplicatedStorage:WaitForChild(\"{}_UNRELIABLE\")", self.config.remote_scope));
+		self.push("\n");
+		self.push_line(&format!("assert(reliable:IsA(\"RemoteEvent\"), \"Expected {}_RELIABLE to be a RemoteEvent\")", self.config.remote_scope));
+		self.push_line(&format!("assert(unreliable:IsA(\"RemoteEvent\"), \"Expected {}_UNRELIABLE to be an UnreliableRemoteEvent\")", self.config.remote_scope));
```
sasial-dev:
```suggestion
		self.push_line(&format!("assert(unreliable:IsA(\"UnreliableRemoteEvent\"), \"Expected {}_UNRELIABLE to be an UnreliableRemoteEvent\")", self.config.remote_scope));
```

docs/.vitepress/components/Editor.vue
```diff
@@ -114,7 +114,7 @@ const beforeMount = (monaco: Monaco) => {
 
 	const Calls = ["SingleSync", "SingleAsync", "ManySync", "ManyAsync"] as const;
 
-	const Options = ["typescript", "write_checks", "casing", "server_output", "client_output", "manual_event_loop", "yield_type", "async_lib"] as const;
+	const Options = ["write_checks", "typescript", "typescript_max_tuple_length", "manual_event_loop", "remote_scope", "server_output", "client_output", "casing", "yield_type", "async_lib"] as const;
```
sasial-dev, line 117: Please add the options to L151 of this file.

sasial-dev (changes requested):

sasial-dev (approved):
LGTM! Thanks for your patience with this one.

PR #108, opened 2024-06-26
Use RunService.PostSimulation instead of RunService.Heartbeat

gaymeowing:
This change provides no benefit as both events are the same, except that the event loop wont show under Heartbeat with robloxs core scripts and instead under PostSimulation in the dev console ScriptProfiler. As its nice to be able to easily tell whats code you can touch within the profiler, and whats code you cant touch.

### DIFF
--- a/zap/src/output/luau/client.rs
+++ b/zap/src/output/luau/client.rs
@@ -133,7 +133,7 @@ impl<'src> ClientOutput<'src> {
 			self.push_line(&format!("local function {send_events}()"));
 			self.indent();
 		} else {
-			self.push_line("RunService.Heartbeat:Connect(function(dt)");
+			self.push_line("RunService.PostSimulation:Connect(function(dt)");
 			self.indent();
 			self.push_line("time += dt");
 			self.push("\n");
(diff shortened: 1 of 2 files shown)


sasial-dev:
Maybe we use this opportunity now that roblox has up to 240 FPS to remove the `1/61` stuff? I'll let that be @jackdotink's call.

jackhexed:
I think removing the limit is a good idea. I don't think that moving to `PostSimulation` would make any difference though. I could see it when benchmarking, but in that case it would be better to have an option to emit microprofiler labels for zap.

sasial-dev:
You'll need to fix the merge conflicts for this PR. I'm not sure about this though - as I run code on `Heartbeat`, and zap would then be a frame behind?

sasial-dev:
I think I'll close this as this is considered a breaking change to move events which is not the intent with `0.6.x`. You're more then welcome to PR in the ability to add an option to emit labels for zap on the microprofiler though! Or of course, you can use the `manual_event_loop` option to run your zap events on `RunService.PostSimulation`.

PR #109, opened 2024-06-27
Always export SendEvents

Ketasaja:
It's common to have long-living `Instance`s that are passed through the network, but that can be deleted, like maps, vehicles, or enemies. Because of event batching, the server firing a Zap event referencing an `Instance` and then deleting it can result in the event trigger being dropped because the `Instance` no longer exists.

While optional `Instance`s exist in Zap, and are fine for things like particles on weapons, they're not suitable for cases like this:
1. A round in a game ends.
2. The server fires an event to all clients telling them this.
3. The server deletes the map.
4. Because deletion replicated but the Zap event trigger didn't yet, other client code doesn't know that the round has ended, and continues assuming the map exists, causing game-breaking errors.

The client can't simply end the round itself if it notices the map is unloaded, either. It doesn't know who won, what the score was, why the round ended, or any other information that any round-ending code would expect.

Pros:
1. Default options don't lead to incorrect or very awkward code. Developers can trust Zap to provide a reliable interface for games of any type or scale. What's intuitive works. Developers can write code with the same order guarantees they'd get from normal `RemoteEvent`s.
6. `manual_event_loop = true` works exactly the same as it did.

Costs:
1. A few extra lines of generated output.
7. One extra export.
8. Developers could wastefully call the function, ignoring the new documentation comment that implies there's an automatic alternative.
sendevents zap


zap/src/output/luau/server.rs
```diff
@@ -861,11 +851,10 @@ impl<'a> ServerOutput<'a> {
 		self.push_line("local returns = table.freeze({");
 		self.indent();
 
-		if self.config.manual_event_loop {
-			let send_events = self.config.casing.with("SendEvents", "sendEvents", "send_events");
+		let send_events = self.config.casing.with("SendEvents", "sendEvents", "send_events");
 
-			self.push_line(&format!("{send_events} = {send_events},"));
-		}
+		self.push_line(&format!("--- Manually flush the outgoing queue."));
```
sasial-dev: Let's remove the comment here.

sasial-dev (changes requested):
Please fix the merge conflicts.

zap/src/output/luau/client.rs
```diff
@@ -812,11 +812,10 @@ impl<'src> ClientOutput<'src> {
 		self.push_line("local returns = table.freeze({");
 		self.indent();
 
-		if self.config.manual_event_loop {
-			let send_events = self.config.casing.with("SendEvents", "sendEvents", "send_events");
+		let send_events = self.config.casing.with("SendEvents", "sendEvents", "send_events");
 
-			self.push_line(&format!("{send_events} = {send_events},"));
-		}
+		self.push_line(&format!("--- Manually flush the outgoing queue."));
```
sasial-dev: Remove this comment.

zap/src/output/luau/client.rs
```diff
@@ -158,10 +139,29 @@ impl<'src> ClientOutput<'src> {
...
+
+		if !self.config.manual_event_loop {
+			self.push_line("local time = 0\n");
+			self.push_line("RunService.Heartbeat:Connect(function(dt)");
+			self.indent();
+			self.push_line("time += dt");
+			self.push("\n");
+			self.push_line("if time >= (1 / 61) then");
+			self.indent();
+			self.push_line("time -= (1 / 61)");
```
sasial-dev: Let's get rid of this and the `time` variable now that roblox supports up to 240 FPS on windows.

zap/src/output/luau/server.rs
```diff
@@ -58,11 +58,9 @@ impl<'a> ServerOutput<'a> {
 		let set_callback = self.config.casing.with("SetCallback", "setCallback", "set_callback");
 		let on = self.config.casing.with("On", "on", "on");
 
-		if self.config.manual_event_loop {
-			let send_events = self.config.casing.with("SendEvents", "sendEvents", "send_events");
+		let send_events = self.config.casing.with("SendEvents", "sendEvents", "send_events");
 
-			self.push_line(&format!("{send_events} = noop,"));
-		}
+		self.push_line(&format!("{send_events} = noop,"));
```
sasial-dev, line 63: Make sure to add the `{send_events}` noop on the client side.

sasial-dev (changes requested):

sasial-dev (approved):
LGTM!

PR #111, opened 2024-06-28
(typescript) Use  `Map<K, V>` to allow any key type 

vnnh:
Currently, the map type outputs `{ [index: K]: V }`, which only allows for `string | number | symbol` key types. Using the `Map<K, V>` type would allow for other key types to be used, like Instances.


sasial-dev (approved):
LGTM!

PR #117, opened 2024-07-01
Add sets

### LINKED ISSUE #83: [FEAT] Set type
## Describe your feature
A set type would allow for sending dictionaries where all the values are `true`. This can be optimized over a map since only the keys have to be sent.

## Alternatives
Literal types like `true` or `false` could be added. Then a `map { [string]: true }` could be used instead assuming zap would be smart enough to optimize for this.

Ketasaja:
Closes #83.

### DIFF
--- a/zap/src/irgen/des.rs
+++ b/zap/src/irgen/des.rs
@@ -197,6 +197,7 @@ impl Des {
 				});
 
 				let (key_name, key_expr) = self.add_occurrence("key");
+				self.push_local(key_name.clone(), None);
 
 				self.push_ty(key, Var::Name(key_name.clone()));
 
--- a/zap/src/irgen/des.rs
+++ b/zap/src/irgen/des.rs
@@ -187,6 +187,24 @@ impl Des {
 				self.push_stmt(Stmt::End);
 			}
 
+			Ty::Set(key) => {
+				self.push_assign(into.clone(), Expr::EmptyTable);
+
+				self.push_stmt(Stmt::NumFor {
+					var: "_".into(),
+					from: 1.0.into(),
+					to: self.readu16(),
+				});
+
+				let (key_name, key_expr) = self.add_occurrence("key");
+
+				self.push_ty(key, Var::Name(key_name.clone()));
+
+				self.push_assign(into.clone().eindex(key_expr.clone()), Expr::True);
+
+				self.push_stmt(Stmt::End);
+			}
+
 			Ty::Opt(ty) => {
 				self.push_stmt(Stmt::If(self.readu8().eq(1.0.into())));
 
--- a/zap/src/irgen/ser.rs
+++ b/zap/src/irgen/ser.rs
@@ -221,7 +221,7 @@ impl Ser {
 				self.push_ty(key, key_name.as_str().into());
 
 				self.push_stmt(Stmt::End);
-				
+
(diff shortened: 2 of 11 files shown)


zap/src/irgen/ser.rs
```diff
@@ -201,6 +201,34 @@ impl Ser {
...
 
+			Ty::Set(key) => {
+				let (len_name, len_expr) = self.add_occurrence("len");
+				let (len_pos_name, len_pos_expr) = self.add_occurrence("len_pos");
+
+				self.push_local(len_pos_name.clone(), Some(Var::from("alloc").call(vec![2.0.into()])));
+				self.push_local(len_name.clone(), Some(0.0.into()));
+
+				let (key_name, _) = self.add_occurrence("k");
+				let (val_name, _) = self.add_occurrence("_");
```
sasial-dev, line 212: Does `_` get picked up for shadowing and need an `add_occurrence`?

zap/src/output/typescript/mod.rs
```diff
@@ -119,6 +119,12 @@ pub trait Output: ConfigProvider {
 				self.push(">");
 			}
 
+			Ty::Set(key) => {
+				self.push("Map<");
```
sasial-dev:
```suggestion
				self.push("Set<");
```

zap/src/output/typescript/mod.rs
```diff
@@ -119,6 +119,12 @@ pub trait Output: ConfigProvider {
...
+				self.push_ty(key);
+				self.push(", true>");
```
sasial-dev:
```suggestion
				self.push(">");
```

sasial-dev (changes requested):
Just these minor things otherwise the code LGTM.

zap/src/irgen/ser.rs
sasial-dev: Maybe it is good practice to keep it with `add_occurrence`? Curious to know what you think.

Ketasaja: It becomes `__1`. Selene doesn't care about shadowing `_` but you're not writing Luau normally so there is no warning if you use an `_` variable. I'm fine with whichever, I could also add string splitting back to emit `_1` if a variable ends in `_`.

sasial-dev: Let's keep it as is then. No need to make it emit `_1` I reckon.

zap/src/irgen/des.rs
```diff
@@ -187,6 +187,24 @@ impl Des {
...
+
+				self.push_stmt(Stmt::NumFor {
+					var: "_".into(),
+					from: 1.0.into(),
+					to: self.readu16(),
+				});
+
+				let (key_name, key_expr) = self.add_occurrence("key");
+
+				self.push_ty(key, Var::Name(key_name.clone()));
```
sasial-dev: `key_1` isn't defined first with
`local key_1`, and so the output looks like this:
```lua
			local value
			value = {}
			for _ = 1, buffer.readu16(incoming_buff, read(2)) do
				key_1 = buffer.readu8(incoming_buff, read(1))
				value[key_1] = true
			end
```

sasial-dev (changes requested):

sasial-dev (approved):
LGTM!

PR #119, opened 2024-07-02
Mention table keys in documentation

Ketasaja:
Why is this worth mentioning? Tables are everything to Luau, `RemoteEvent` serialization doesn't respect them, and Zap does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Zap is a blazingly fast networking solution for Roblox.
 
 - Zap packs data into buffers with no overhead. The same data can be sent using a fraction of the bandwidth.
 - Zap doesn't compromise on performance. Zap's packing and unpacking is typically faster than Roblox's generic encoding.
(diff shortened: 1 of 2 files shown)


sasial-dev:
I'm not sure if we want to encourage some of those behaviors tbh. Things like holey arrays are bad practice.

Ketasaja:
I wouldn't encourage holey arrays either, but I think using numbers as hashtable keys is idiomatic (like Roblox user identifiers), and if you're ever sending identifier `1` through a `RemoteEvent`, you accidentally made an "array". As for other cases, I think using `string` keys for everything can cause type safety issues developers may be looking to avoid.

I'm fine with this PR being rejected though.

jackhexed:
I'm fine with this general idea being listed in the docs, but I would rephrase it to something like "Extended Type Support" and "Zap supports far more types than Remote serialization does. Maps with numbers and instances as keys is a good example."

sasial-dev (changes requested):
Just following this up! :)
Please review @jackdotink's comments

sasial-dev (approved):
LGTM!

PR #120, opened 2024-07-07
Allow boolean literals in enums

gaymeowing:
This pr adds support for boolean literals in enums, a pattern that is useful for result types

ex:
```Luau
event EVENTI_RESULT = {
	from: Server,
	type: Reliable,
	call: ManyAsync,
	data: enum "succeeded" {
		true {
			from_action: enum { SCHEDULE, END, REMOVE, START, TELEPORT }
		},
		false {
			from_action: enum { SCHEDULE, END, REMOVE, START, TELEPORT },
			error: string,
		}
	},
}
```

### DIFF
--- a/zap/src/irgen/mod.rs
+++ b/zap/src/irgen/mod.rs
@@ -329,6 +329,7 @@ pub enum Expr {
 	Nil,
 
 	// Literals
+	StrOrBool(String),
 	Str(String),
 	Var(Box<Var>),
 	Num(f64),
@@ -441,6 +442,15 @@ impl Display for Expr {
 			Self::True => write!(f, "true"),
 			Self::Nil => write!(f, "nil"),
 
+			Self::StrOrBool(string) => {
+				if string == "false" || string == "False" {
+					return write!(f, "false")
+				} else if string == "true" || string == "True" {
+					return write!(f, "true")
+				} else {
+					return write!(f, "\"{}\"", string)
+				}
+			},
 			Self::Str(string) => write!(f, "\"{}\"", string),
 			Self::Var(var) => write!(f, "{}", var),
 			Self::Num(num) => write!(f, "{}", num),
--- a/zap/src/irgen/mod.rs
+++ b/zap/src/irgen/mod.rs
@@ -445,7 +445,7 @@ impl Display for Expr {
 			Self::StrOrBool(string) => {
 				if string == "false" || string == "False" {
 					return write!(f, "false")
-				} else if string == "true" || string == "True" {
+				} else if string == "true" {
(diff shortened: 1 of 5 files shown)


sasial-dev:
I'm still unconvinced with this PR, can you explain to me more around the motivation of this PR and why you can't persue the current alternatives?

gaymeowing:
> I'm still unconvinced with this PR, can you explain to me more around the motivation of this PR and why you can't persue the current alternatives?

The motivation is that it'd be an easy way to support result types for usecases of reporting datastore errors, roblox api errors, ect, to the client. Plus this is more intended to be a quick fix, as I'd like to not have to use "false" and "true" strings. That don't even atleast coerce into booleans in luau. And this can be replaced with a better syntax in the rewrite, and I'd like to not fully properly implement a solution as this can be just made better in the rewrite.

sasial-dev:
Just out of curiosity, what syntax would you suggest in the rewrite?

zap/src/irgen/mod.rs
```diff
@@ -441,6 +442,15 @@ impl Display for Expr {
 			Self::True => write!(f, "true"),
 			Self::Nil => write!(f, "nil"),
 
+			Self::StrOrBool(string) => {
+				if string == "false" || string == "False" {
```
sasial-dev: I think this is a gotcha. Let's make it more explicit so that people can still use `False` if they want a string.
Or, we can put this behind a setting. Up to you.

If you go with the first option:
```suggestion
				if string == "false" {
```

zap/src/irgen/mod.rs
```diff
@@ -441,6 +442,15 @@ impl Display for Expr {
...
+					return write!(f, "false")
+				} else if string == "true" || string == "True" {
```
sasial-dev: See above - for the first option:
```suggestion
				} else if string == "true" {
```

sasial-dev (changes requested):

gaymeowing:
Probably a result type, something like
```
result<value_on_sucess, value_on_fail>
```
that'd then serialize into
```luau
type MyResult = {
     success: true,
     -- whatever is included in value_on_sucess, if value_on_sucess is just like a singular type ie u8 or u8 then it will create a value field
} | {
   success: false,
   -- same as the success side but using value_on_fail
}
```

jackhexed:
result types like that will not be happening in rewrite

sasial-dev (changes requested):
Your PR doesn't fix the typings emitted in both lua and TS. They are `"true"` and `"false"` atm

sasial-dev (approved):
LGTM!

PR #121, opened 2024-07-21
Remove function shadowing

Ketasaja:
Reported by @Crystalflxme in [ROSS](https://discord.gg/qpgNftpnpf).


Crystalflxme:
Thanks for bringing this up! One thing that was mentioned back when I originally noticed the shadowing was that avoiding the usage of upvalues as it was doing is apparently faster. In that case, it's probably best to change the variable name used in the spawned function.

jackhexed:
beat me to it @Crystalflxme

sasial-dev (approved):
LGTM!

PR #124, opened 2024-07-27
Add function return types to client emit

Ketasaja:
[Reported](https://discord.com/channels/385151591524597761/1064606584430411906/1266472145559486494) by @crystalflxme in [ROSS](https://discord.gg/qpgNftpnpf).

PR #125, opened 2024-07-27
Add function returns with `YieldType::Future` or `YieldType::Yield`

Ketasaja:
Another attempt of #124. Emit is now better when using `YieldType::Future` or `YieldType::Yield`, and no worse when not.

Assumes `Future` type exists in `async_lib` when `async_type = "future"`.

There's no type when using `Promise`s without TypeScript, this can probably change once the new solver releases, but it would be breaking.


sasial-dev (approved):
LGTM but will get @jackdotink to review

jackhexed (approved):
lgtm

PR #128, opened 2024-08-04
Remove returning `table.freeze`

Ketasaja:
`table.freeze` inconsistently causes firing events with Zap not to typecheck at all in the old solver. From testing it's when `table.freeze` is combined with `--!nocheck` or `--!nonstrict` and called into from a `--!strict` module.

`table.freeze` is bugged in the new solver, and even if that were fixed the new solver isn't yet quite ready for general use.

It should be added back later, which would be easy.

### DIFF
--- a/zap/src/output/luau/client.rs
+++ b/zap/src/output/luau/client.rs
@@ -559,13 +559,13 @@ impl<'src> ClientOutput<'src> {
 			.iter()
 			.filter(|ev_decl| ev_decl.from == EvSource::Client)
 		{
-			self.push_line(&format!("{name} = table.freeze({{", name = ev.name));
+			self.push_line(&format!("{name} = {{", name = ev.name));
 			self.indent();
 
 			self.push_return_fire(ev);
 
 			self.dedent();
-			self.push_line("}),");
+			self.push_line("},");
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


sasial-dev:
I'm not convinced about this. It seemed to be working fine for me in Luau LSP? Do you find this issue in VSCode or only in studio?

I also would like @kalrnlo's opinion as they originally PR'ed in #97.

gaymeowing:
I'm ok if it gets removed for rn tbh

Ketasaja:
VS Code, Luau-LSP 1.32.1. Luau has issues specifically with calls from strict into nonstrict mode. I use `table.freeze` in all my code (I use old solver and strict mode everywhere), `table.freeze` is correct to return from Zap, my PR is meant to be temporary.

```luau
--!strict
local network_client = require(script.network_client)

network_client.test.fire({}) -- no error unless `table.freeze` removed, or generated output is --!strict instead of --!nocheck
```
```
opt casing = "snake_case"
opt client_output = "src/client/network_client.luau"
opt server_output = "src/server/network_server.luau"

event test = {
	from: Client,
	type: Reliable,
	call: SingleSync,
	data: f64,
}
```

sasial-dev (approved):
LGTM!

PR #129, opened 2024-08-10
Add Generalised Tooling Support

### LINKED ISSUE #127: [FEAT] Provide support for Packet Profiler
Hi!

After a discussion in the packet profiler thread in the Roblox OSS server, I'd like to request zap adding support for the packet profiler plugin. As zap uses the one-remote schema, reading `RemoteObject.Name` becomes moot for the profiler.
In consideration of this, packet profiler provides support for custom naming of remotes. You can find the description detailed on the DevForum thread:
https://devforum.roblox.com/t/version-200-packet-profiler-easily-measure-all-of-your-remotes/1890924#support-for-remote-compression-libraries-8

sasial-dev:
~Blocked by https://github.com/Pyseph/PacketProfiler/pull/8~

cc @Pyseph

For tools such as PacketProfiler, there is a need for ability to deserialise a zap buffer while 'in transit' for debugging.
This PR aims to create a standardised way to do this, accepting a function with all the remote's parameters and returning a table with the name and arguments such as:
```lua
			table.insert(events, {
				Name = "RemoteName",
				Arguments = { value }
			})
```


zap/src/output/tooling.rs
```diff
@@ -0,0 +1,347 @@
...
+		self.push_line("incoming_read = incoming_read + len");
+		self.push("\n");
+		self.push_line("return pos");
+
+		self.dedent();
+		self.push_line("end");
+		self.push_line("local len = buffer.len(incoming_buff)");
+		self.push("\n");
+
+		self.push_tydecls();
```
sasial-dev, line 344: One of the major intents was not to touch the rest of the codebase (e.g `irgen`), and that meant that the types functions have to be inside the function. Let me know if you think this is an issue.

sasial-dev: RobloxStudio-njRe63Fu@2x

This doesn't seem to be as slow as I was expecting it to be...

jackhexed (approved):

zap/src/output/tooling.rs
sasial-dev: @jackdotink - before I merge can you confirm if this is ok?

jackhexed: lgtm

PR #131, opened 2024-08-14
Fix sync functions

Ketasaja:


sasial-dev (approved):
Thank you for the speedy fix!

PR #136, opened 2024-09-24
Add polling

Ketasaja:
- New option `call_default`.
- New `EvCall:Polling` variant.
- Explicit iteration exists because type inference for loop variables with generalized iteration isn't stable yet until the new solver releases.


docs/config/options.md
```diff
@@ -37,6 +37,22 @@ The paths are relative to the configuration file and should point to a lua(u) fi
...
+
+<CodeBlock code = 'opt event_handling = "signal"' />
+
+### Example
+<script setup lang="luau">
+-- Example of using a Zap polling event on the server.
+for player, payload in zap.my_event do
+	print(player, payload)
+end
+</script>
```
sasial-dev: `<script setup` is actually a `vue` thing where it runs code at compile time.
Up on ~ L20 add something like
```ts
const handleExample = `-- Example of using a Zap polling event on the server.
for player, payload in zap.my_event do
	print(player, payload)
end`
```
and then here you can replace it with
```suggestion

<CodeBlock :code="handleExample" />
```

zap/src/output/luau/client.rs
```diff
@@ -1,5 +1,7 @@
+use std::collections::binary_heap::PeekMut;
+
```
sasial-dev: Is this an accidental autoimport?

sasial-dev (changes requested):
Thanks for this!
I started by reviewing general code stuff. I'll let @jackdotink review first and then I'll go through and test the output locally.

Ketasaja:
Instead of adding a new event field I added `EvCall::Polling`.

I need to make the global option configure the default `EvCall`, update the docs, fix TypeScript polling, and add error handling for when there's no `call` specified.

Ketasaja:
I think in a rewrite or in the future in general, Zap should report an error and exit immediately when one is encountered.

sasial-dev:
The reason why we did that is for error recovery I think. It would make it much eaisier for stuff like building an LSP etc

zap/src/config.rs
```diff
@@ -37,6 +37,12 @@ impl<'src> Config<'src> {
 	}
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum EventHandling {
+	Polling,
+	Signal,
+}
+
```
jackhexed: why do we need this and `EvCall::Polling`

zap/src/output/luau/client.rs
```diff
@@ -199,62 +201,73 @@ impl<'src> ClientOutput<'src> {
...
-			self.push_line(&format!("if events[{id}] then"));
-		} else {
-			self.push_line(&format!("if events[{id}][1] then"));
-		}
+		match ev.call {
+			EvCall::Polling => {
+				// Event types without data use `true` as a placeholder.
+				self.push_line(&format!(
+					"table.insert(polling_payload_queues[{id}], if value == nil then true else value)"
+				));
```
jackhexed: if true is being used as a placeholder, then what happens to real true values? I recommend we use some "null" userdata to represent nil

zap/src/output/luau/client.rs
```diff
@@ -691,6 +705,21 @@ impl<'src> ClientOutput<'src> {
...
+		let iter = self.config.casing.with("Iter", "iter", "iter");
+		let id = ev.id;
+		self.push_indent();
+		self.push(&format!("{iter} = polling_iterators[{id}] :: () -> "));
+		if let Some(data) = &ev.data {
+			self.push("(() -> ");
+			self.push_ty(&data);
+			self.push(")");
+		} else {
+			self.push("(() -> true)")
```
jackhexed: again with the true stuff

zap/src/output/luau/client.rs
```diff
@@ -701,9 +730,14 @@ impl<'src> ClientOutput<'src> {
...
 
+			if let EvCall::Polling = ev.call {
+				self.push_explicit_iter(ev);
+			}
+
 			match ev.call {
 				EvCall::SingleSync | EvCall::SingleAsync => self.push_return_setcallback(ev),
 				EvCall::ManySync | EvCall::ManyAsync => self.push_return_on(ev),
+				_ => (),
 			}
```
jackhexed: why is this `if let` above the match?

zap/src/output/luau/server.rs
```diff
@@ -211,23 +213,32 @@ impl<'a> ServerOutput<'a> {
 			self.push_stmts(&des::gen(data, "value", true));
 		}
 
-		if ev.call == EvCall::SingleSync || ev.call == EvCall::SingleAsync {
-			self.push_line(&format!("if events[{id}] then"))
+		if let EvCall::Polling = ev.call {
+			self.push_line(&format!("table.insert(polling_player_queues[{id}], player)"));
+			// Event types without data use `true` as a placeholder.
+			self.push_line(&format!(
+				"table.insert(polling_payload_queues[{id}], if value == nil then true else value)"
```
jackhexed: again with true

zap/src/output/luau/server.rs
```diff
@@ -838,19 +850,39 @@ impl<'a> ServerOutput<'a> {
...
+		let iter = self.config.casing.with("Iter", "iter", "iter");
+		let id = ev.id;
+		self.push_indent();
+		self.push(&format!("{iter} = polling_iterators[{id}] :: () -> "));
+		if let Some(data) = &ev.data {
+			self.push("(() -> (Player, ");
+			self.push_ty(&data);
+			self.push("))");
+		} else {
+			self.push("(() -> (Player, true)")
```
jackhexed: again with true

jackhexed (changes requested):

zap/src/output/luau/client.rs
sasial-dev: Or maybe we use a counter for events w/o data instead of an array - sorta like how it works already for queuing

Ketasaja: > Or maybe we use a counter for events w/o data instead of an array - sorta like how it works already for queuing

Something needs to be returned to the loop or iteration will end, there's no player to return on the client.

Ketasaja: > if true is being used as a placeholder, then what happens to real true values? I recommend we use some "null" userdata to represent nil

I don't see much benefit to this unless you assign a metatable that errors on everything, but you don't like metatables. I could make this change though.

```luau
for meaningless_value in event do
    if meaningless_value then
        -- Same behavior either way.
    end
end
```

Ketasaja: You'd also need to lie to the type system because `newproxy(true)` returns `any`, which would be worse than `true`.

Ketasaja:
I changed the placeholder value to an explanatory string literal because I don't think a `newproxy` userdata adds any benefit. `newproxy` values are `any` type, which is worse than a singleton. A metatable that errors on everything might be useful, but Zap doesn't like metatables.
```luau
for meaningless_value in event do
    if meaningless_value then
        -- Same behavior whether `meaningless_value` is a `newproxy` userdata or not.
    end
end
```

I can change this if necessary.

zap/src/output/luau/client.rs
jackhexed: we need to do something because if we use `true` then `boolean?` becomes impossible

sasial-dev:
Does it matter that `newproxy` is typed as any? We can type it as `never` ourselves if we really want to.

sasial-dev:
I'd also prefer we didn't use a string singleton.

Ketasaja:
> Does it matter that `newproxy` is typed as any? We can type it as `never` ourselves if we really want to.

`never` won't cause type errors. A string singleton will at least cause type errors in strict mode whenever it's used in a non-string or other string literal context, and would be obvious once inspected at runtime. If the userdata is passed around it won't be obvious what it is outside of networking code when debugging.

If you want the change to userdata with that said, I'll make it, but I think a string singleton is the least wrong option.

sasial-dev:
*`unknown`, not `never` sorry

sasial-dev:
As discussed with @jackdotink - I think this PR is overcomplicated and has a ton of merge conflicts. We'd still be open to a new implimentation, though!

PR #137, opened 2024-09-27
(typescript) Add ` | undefined` to emit for optional

vnnh:
This makes optional properties emit as `{ property?: type | undefined }` which fixes issues when `exactOptionalPropertyTypes` is set to true in tsconfig, such as when trying to fire an event with a possibly undefined value.


sasial-dev (approved):
Thank you!

PR #141, opened 2024-11-30
Named parameters

nezuo:
TODO:
- [x] Docs

### DIFF
--- a/docs/config/functions.md
+++ b/docs/config/functions.md
@@ -1,17 +1,27 @@
 <script setup lang="ts">
 const example = `funct Test = {
     call: Async,
-    args: struct {
-        foo: u8,
-        bar: string
-    },
+    args: (Foo: u8, Bar: string),
     rets: enum { Success, Fail }
 }`
 
-const multiArgsRetsExample = `funct MultipleArgsRets = {
+const argsExample = `funct OneUnnamedParameter = {
     call: Async,
-    args: (boolean, u8),
-    rets: (boolean, string)
+    args: u8,
+    rets: enum { Success, Fail }
+}
+
+funct TwoUnnamedParameters = {
+    call: Async,
+    args: (u8, string),
+    rets: enum { Success, Fail }
+}
+`
+
+const multipleRets = `funct MultipleRets = {
+    call: Async,
+    args: boolean,
+    rets: (enum { Success, Fail }, string)
 }`
 </script>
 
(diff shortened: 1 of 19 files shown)


zap/src/output/luau/client.rs
```diff
@@ -210,12 +215,16 @@ impl<'src> ClientOutput<'src> {
 
 		self.indent();
 
-		let values = self.get_values(&ev.data);
+		let values = self.get_values(ev.data.len());
```
sasial-dev, line 218: I know we don't care about the value names in the generated code, but if we're going to the bother of writing a function I'm curious to know why you didn't just use them here?

zap/src/parser/convert.rs
```diff
@@ -319,17 +321,24 @@ impl<'src> Converter<'src> {
...
-		let data = evdecl
-			.data
-			.as_ref()
-			.map(|types| types.iter().map(|ty| self.ty(ty)).collect::<Vec<_>>());
+		let data = evdecl.data.as_ref().map(|parameters| {
+			parameters
+				.parameters
+				.iter()
+				.map(|(identifier, ty)| {
+					let name = identifier.map(|identifier| identifier.name);
```
sasial-dev, line 350: Are parms with the same name considered okay? Do we need to error for them? If so that would explain my first comment.

zap/src/parser/grammar.lalrpop
```diff
@@ -121,6 +121,22 @@ TyList: Vec<SyntaxTy<'input>> = {
     "(" <tys:Comma<Ty>> ")" => tys,
 }
 
+Paremeters: SyntaxParameters<'input> = {
```
sasial-dev: nit: this is spelt wrong

sasial-dev (changes requested):
Looking good! Got a few initial questions

zap/src/output/luau/client.rs
nezuo: I didn't do it just in case the user enters a name that conflicts with other variables in the output.

zap/src/parser/convert.rs
nezuo: If you think we need an error I can add it. I didn't because it should be pretty obvious you can't have duplicate parameter names.

sasial-dev: I think we should add an error, yeah.

nezuo:
Added error for duplicate parameters:
image

docs/config/functions.md
```diff
@@ -48,15 +58,16 @@ Use synchronous functions with extreme caution.
...
 
-<CodeBlock :code="multiArgsRetsExample" />
+<CodeBlock :code="argsExample" />
 
 ### `rets`
 
 This field determines the data that is sent back to the client from the server. It can be any [Zap type](./types.md).
 
 - If the server doesn't return any data, the `rets` field should be excluded.
+- Unlike `args`, `rets` cannot be named.
```
sasial-dev, line 70: Maybe we need a better error (sorry) then
```
error[2002]: expected ")", ",", "?", "["
  ┌─ net.zap:4:16
  │
4 │     rets: (test: enum { Success, Fail }, bar: string)
  │                ^ unexpected token
```

sasial-dev (approved):
Other than the comment above - LGTM

nezuo:
Improved error for named rets:
```lua
error[3016]: rets cannot be named
   ┌─ net.zap:88:12
   │
88 │     rets: (foo: f64, boolean),
   │            ^^^ must be removed
```

sasial-dev:
Thank you!

PR #142, opened 2024-12-04
Docs: Add API documentation for functions

Blizzero:
- Renamed "Event Usage" to "Generated API".
- Reformatted page and added documentation for functions.


docs/.vitepress/config.mts
```diff
@@ -26,7 +26,7 @@ const sidebar = [
     text: 'Using Zap',
     items: [
       { text: 'Generate Code', link: '/usage/generation' },
-      { text: 'Event Usage', link: '/usage/events' },
+      { text: 'Generated API', link: '/usage/generated-api' },
```
sasial-dev: I'm not convinced on `Generated API`. Maybe something like `Using Zap`?

docs/usage/generated-api.md
```diff
@@ -0,0 +1,199 @@
+<script setup lang="ts">
+const eventConfigFile = `event MyEvent = {
+    from: Server,
+    type: Reliable,
+    call: ManyAsync,
+    data: (Options: struct {
+        foo: string,
+        bar: u8,
+    }),
```
sasial-dev:
```suggestion
    data: struct {
        foo: string,
        bar: u8,
    }
```

docs/usage/generated-api.md
```diff
@@ -0,0 +1,199 @@
...
+    data: (Foo: boolean, Bar: u8)
+}
+`
+
+const functConfigFile = `funct GetScore = {
+    call: Async,
+    args: struct {
+        roundId: string,
+        category: enum { HighScore, LowScore, AverageScore },
+    },
```
sasial-dev:
```suggestion
    args: (roundId: string, category: enum { HighScore, LowScore, AverageScore })
```

docs/usage/generated-api.md
```diff
@@ -0,0 +1,199 @@
...
+    rets: u16,
+}
+`
+</script>
+
+# Using Your Generated Code
+
+After [generating code](./generation.md), you will gain access to Zap's generated API. While this API is fully typesafe, meaning it will not throw errors for invalid usage, your static analysis tooling will. To solve this:
```
sasial-dev: This sentence doesn't make much sense.

docs/usage/generated-api.md
```diff
@@ -0,0 +1,199 @@
...
+```lua
+Zap.MyEvent.Fire(Player, {
+    foo = "baz",
+    bar = 1,
+})
+```
+
+### FireAll
+
+The `FireAll` function takes the event's data as its arguments. It will fire the event to all players.
```
sasial-dev, line 110: Maybe we should note this function can be disabled? See https://zap.redblox.dev/config/options.html#disable-fire-all for more info.

sasial-dev (changes requested):
LGTM - I've just made a few nit picky comments.

docs/.vitepress/config.mts
Blizzero: I'm also not particularly convinced on it, but I couldn't really think of any good alternatives. `Using Zap` is the name of the parent section, so that seems even stranger. Maybe "Event and Function API"?

docs/usage/generated-api.md
Blizzero: This sentence _is_ valid English, but if we're already getting confused by it, it should probably be changed. What in particular is nonsensical about it? It reads fine to me.

sasial-dev: The way it mentions typesafety is confusing, and it also contradicts the homepage
Brave Browser-cJyK2fA5@2x

As zap _will_ throw errors for invalid data (e.g. think wrong enums, writechecks if enabled) and buffer methods will also force valid usage.

Blizzero: Ah, I see. I pretty much just reformatted the existing sentence: "While Zap will not throw errors for invalid usage, your static analysis tooling will."

Should be easy enough to rewrite appropriately.

docs/.vitepress/config.mts
sasial-dev: Or maybe "Sending Events and Functions"

Blizzero: Though it's implied, that ignores the fact half the page is about "receiving" events and functions.

sasial-dev: Ok - let's do "Event and Function API" then

docs/usage/generated-api.md
Blizzero: I'm thinking of just removing that blurb and the proceeding 2 bullet points. Anyone using Zap is probably already aware of those options, and they aren't really Zap specific for us to even address. Thoughts?

sasial-dev: Yeah - I'd hope they'd already know too. That's fine with me.

sasial-dev (approved):
Thank you!

PR #143, opened 2024-12-31
Optimize event ids by splitting them up by server/client and reliable/unreliable

nezuo:
Currently, all events and functions share event ids in the same pool. This is unnecessary and optimizing it will allow for more events/functions before needing to increase the bytes required for the event id.

To give an example, reliable events and functions are never sent through the unreliable remote. This means we can have separate pools of ids for reliable and unreliable.


nezuo:
Event ids used to start at `1` instead of `0`. I'm not sure why this was the case, let me know if I'm missing something!

zap/src/config.rs
```diff
@@ -31,9 +31,55 @@ pub struct Config<'src> {
...
+
+		reliable_count + self.fndecls.len()
+	}
+
+	pub fn client_unreliable_count(&self) -> usize {
+		self.evdecls
+			.iter()
+			.filter(|evdecl| evdecl.from == EvSource::Server && evdecl.evty == EvType::Unreliable)
+			.count()
+	}
```
jackhexed, line 67: why are we doing this instead of having separate reliable and unreliable counts?

nezuo: This allows for more events before increasing the ids. Imagine 256 reliable client to server events and 256 reliable server to client events. With how I've done it, that only needs a u8. If we combined them, it would need a u16. It's possible to separate them because the events are one-way.

I might have misunderstood the question though.

jackhexed: Yeah I was asking why not have `server_reliable` and `server_unreliable` counters, and the respective client counters too

jackhexed: I think that would be better than putting them into one list and then iterating over it each time

jackhexed (approved):

sasial-dev:
@nezuo apologies - could you please fix the merge conflicts in your PR?

zap/src/parser/convert.rs
```diff
@@ -12,33 +12,43 @@ use super::{
...
 			match decl {
 				SyntaxDecl::Ty(tydecl) => {
 					tydecls.insert(tydecl.name.name, tydecl.clone());
 				}
 
-				SyntaxDecl::Ev(_) | SyntaxDecl::Fn(_) => ntdecls += 1,
+				SyntaxDecl::Ev(ev_decl) => match ev_decl.from {
+					EvSource::Server => client_unreliable_event_count += 1,
+					EvSource::Client => server_unreliable_event_count += 1,
+				},
```
sasial-dev: This counts for all, not just unreliable IDs

zap/src/parser/convert.rs
```diff
@@ -370,18 +410,23 @@ impl<'src> Converter<'src> {
...
+			if min > max_unreliable_size {
 				self.report(Report::AnalyzeOversizeUnreliable {
 					ev_span: evdecl.span(),
 					ty_span: evdecl.data.as_ref().unwrap().span(),
-					max_size: self.max_unreliable_size,
+					max_size: max_unreliable_size,
 					size: min,
 				});
-			} else if !max.is_some_and(|max| max < self.max_unreliable_size) {
+			} else if max.is_none_or(|max| max >= max_unreliable_size) {
```
sasial-dev, line 429: Have to double check the behavior here, is the packet dropped at 900 (`>`) or 901 (`>=`)?

sasial-dev (changes requested):
Sorry for the delay in reviewing this! This looks pretty good - I just have a few comments before we can merge this in :)

zap/src/parser/convert.rs
nezuo: This functionality should be equivalent as I used rust-analyzer to automatically fix it. It was being linted by clippy. It seems like it should be `>` but that's not what it was doing previously.

sasial-dev (approved):
Thank you!

PR #146, opened 2025-01-13
Prevent remote function errors with actors

Ketasaja:
Since there can be multiple Zap instances when using actors, there can be multiple `RemoteEvent.OnClientEvent` listeners. Only the listener from the VM that fired the remote function will have a valid thread to resume execution at. This change simply ignores the remote function return on any other VM.

Reported by Phoenix in ROSS.


sasial-dev (commented):
I'm not sure how I feel about your change to be honest, as while it does fix the issue, it _technically_ introduces the possibility for an event to be dropped which I don't like.
So I'll leave it open for a little and get some input from others.

(message from [this Discord thread](https://canary.discord.com/channels/385151591524597761/1064606584430411906/1328243350192652339))

sasial-dev (changes requested):
This needs to properly handle `YieldType::Promise` as well as otherwise we'll still get the same error

sasial-dev (approved):
Thank you!

PR #151, opened 2025-01-16
Document unknown

Ketasaja:
Normally you wouldn't document the type's usage but I think unexperienced Luau users may not know what to do with the value.

### DIFF
--- a/docs/config/types.md
+++ b/docs/config/types.md
@@ -104,22 +104,7 @@ The length of strings can be constrained by placing [a range](#as-a-recap) withi
 
 ## Unknown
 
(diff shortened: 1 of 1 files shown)


sasial-dev:
Thank you for this! I’m not sure if refining the type is the best idea on the docs, as that may push people to use unknown instead of tagged enums.

Maybe it should be something along the lines of when the data types cannot be known at runtime or fulfilled with tagged enums.

Ketasaja:
I updated the description. I could remove the refinements example but tagged enums aren't always possible.

sasial-dev (approved):
Thank you!

PR #152, opened 2025-01-17
Start Tests

sasial-dev:
This pull request starts the tests for `0.6.x`.

Currently, there are the following types of tests:

- [x] Passing the code through Selene (to check for shadowing and other lints we haven't disabled)
- [x] Checking if the code compiles/runs in lune
- [x] Some irgen tests passing data and checking it is the same after being serialised and deserialised*

The first two types are automatically run for every test fixures inside `zap/tests/files`, while the others are manually written.

The location of the `.snap` files etc is subject to change (I may put them all in the same directory instead).

*The only exception to this is that we cannot test CFrame serialisation without polyfilling it in lune which doesn't seem worth it.


sasial-dev:
Brave Browser-cHpj3vXk@2x
A thing to note is that CI is much slower... I wonder if we should change the build script in CI to just run `cargo test` and not `cargo build`.

Ketasaja:
```luau

**Comment by Ketasaja** [comment]
Test case for #153 you asked for.
```luau
local server = require(script.Parent:FindFirstChild("server"))

server.MultipleRets.set_callback(function()
	return "a", "b", true
end)
```
```luau
local client = require(script.Parent:FindFirstChild("client"))

local first, second, third = client.MultipleRets.call(true)
assert(first == "a", "")
assert(second == "b", "")
assert(third == true, "")
```

```
opt casing = "snake_case"
opt client_output = "src/client/client.luau"
opt server_output = "src/server/server.luau"
funct MultipleRets = {
    call: Async,
    args: boolean,
    rets: (string, string, boolean)
}
```

sasial-dev:
^ CI should fail until I merge from `master` back into this branch

zap/Cargo.toml
```diff
@@ -18,3 +18,13 @@ codespan-reporting = "0.11.0"
...
+
+[dev-dependencies]
+anyhow = "1.0"
+full_moon = { version = "1.2.0", features = ["luau"] }
+insta = { version = "1.42.0", features = ["glob"] }
+lune = { git = "https://github.com/lune-org/lune.git", rev = "5d1401cdf611cc9f215f03bdf16ada0e77717edf" }
+selene-lib = "0.28.0"
+serde_yml = "0.0.12"
+tokio = { version = "1", features = ["full"] }
+toml = "0.8.19"
```
sasial-dev, line 30: This looks like a lot but:

- `full_moon` is deduped with `selene`
- `tokio` is deduped with `lune`
- `serde_yml`/`toml` are so that you can run selene (VSC extension or CLI) in your IDE to debug test results :)

polyfill.yml
```diff
@@ -0,0 +1,40 @@
+base: luau
```
sasial-dev, line 1: The polyfill is in the root so that you can run selene with the CLI or VSCode extension if you wish

zap/src/lib.rs
```diff
@@ -1,7 +1,7 @@
-mod config;
-mod irgen;
-mod output;
-mod parser;
+pub mod config;
+pub mod irgen;
+pub mod output;
+pub mod parser;
```
sasial-dev, line 4: `irgen` tests need to access this and I think it's alright to have these public as we don't publish to cargo

zap/tests/irgen/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+async fn test_function() {
+	let (config, reports) = parse(include_str!("../files/function.zap"));
+
+	assert!(config.is_some());
+	assert!(reports.is_empty());
+
+	let default_values: HashMap<&str, Vec<&str>> = HashMap::from([
+		("Test__ARGS", vec!["0", "\"foo\""]),
+		("Test__RETS", vec!["\"Success\"", "\"bar\""]),
+	]);
```
sasial-dev, line 184: Not the happiest with how `default_values` is done in this file. Let me know if you have any better ideas.

zap/tests/selene/checker.rs
```diff
@@ -0,0 +1,52 @@
...
+
+		if let Some(next_base) = base {
+			resolved_libs.append(&mut resolve_standard_library_bases(&next_base))
+		}
+	}
+
+	resolved_libs
+}
+
+pub fn initialise_selene() -> Selene {
```
sasial-dev, line 27: This function mimics the way the CLI loads the config/standard libs which is why it's a bit weird

zap/tests/selene/mod.rs
```diff
@@ -0,0 +1,55 @@
+mod checker;
+
+use std::{borrow::Cow, fs, sync::LazyLock};
+
+use checker::{initialise_selene, Selene};
+use insta::{assert_debug_snapshot, glob, Settings};
+
+static SELENE: LazyLock<Selene> = LazyLock::new(initialise_selene);
```
sasial-dev, line 8: ^ This bumps the MSRV of zap to `1.80.0` which is fine as we compile on `stable` and we aren't a rust crate! But good to be aware of :)

sasial-dev (commented):
Some of my thoughts to help make sense of when reviewing

jackhexed (approved):
lgtm but I'm not sure how qualified I am to review this as I know very little about the dependencies we're bringing in here

sasial-dev:
The deps are only in the test build! So no change in the real bundle size :)

PR #153, opened 2025-01-18
Fix remote function multiple returns

Ketasaja:
Reported by J-ian in ROSS.

This also fixes a separate issue with #143 on latest where this code still refers to `events` instead of `reliable_events`.

I stopped calling `get_unnamed_values` because there are other callsites I didn't want to mess with.

### DIFF
--- a/zap/src/output/luau/server.rs
+++ b/zap/src/output/luau/server.rs
@@ -292,23 +292,18 @@ impl<'a> ServerOutput<'a> {
 			));
 		}
 
-		self.push_line(&format!("if events[{server_id}] then"));
+		self.push_line(&format!("if reliable_events[{server_id}] then"));
 
 		self.indent();
 
-		let rets = if !fndecl.args.is_empty() {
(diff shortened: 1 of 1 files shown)


zap/src/output/luau/server.rs
```diff
@@ -292,23 +292,18 @@ impl<'a> ServerOutput<'a> {
...
-		let rets = if !fndecl.args.is_empty() {
-			(1..=fndecl.args.len())
-				.map(|i| {
-					if i > 1 {
-						format!("rets{}", i)
-					} else {
-						"rets".to_string()
-					}
-				})
+		let rets = fndecl.rets.clone().unwrap_or(vec![]);
```
sasial-dev: Why are we cloning here?

Ketasaja: `error[E0507]: cannot move out of `fndecl.rets` which is behind a shared reference`

sasial-dev (approved):
Thank you!

PR #154, opened 2025-01-21
Pull `var_occurrences` upward

Ketasaja:
Fix for shadowing in nested irgen calls.


sasial-dev (approved):
Thank you for fixing my code - of course it's a dead simple fix...

PR #157, opened 2025-01-23
Fix Tooling Bugs

sasial-dev:
This fixes two tooling regressions introduced by #141 (not defining tuples correctly) and #143 (adding `else` without an `if` above).


jackhexed (approved):
lgtm

PR #159, opened 2025-01-24
Add vectors

Ketasaja:
Adds [faster](https://x.com/sleitnick/status/1882879057953804344) `vector` type and keyword. `Vector3` and `vector` are perfectly compatible at runtime but have different types.

Future change could be for `Vector` to store `NumTy`s instead of `Ty`s but `NumTy` doesn't have ranges right now. Alternative was `Vector<Box<NumTy>, Box<NumTy>, Option<Box<NumTy>, `Option<Range>, Option<Range>, Option<Range>>` which I didn't like.


zap/src/config.rs
```diff
@@ -291,6 +292,50 @@ impl<'src> Ty<'src> {
...
+						NumTy::U8 => 1,
+						NumTy::U16 => 2,
+						NumTy::U32 => 3,
+						NumTy::I8 => 1,
+						NumTy::I16 => 2,
+						NumTy::I32 => 4,
+						NumTy::F32 => 4,
+						NumTy::F64 => 8,
+					},
+					_ => panic!(),
```
sasial-dev: Shouldn't this be `unreachable!` instead?

zap/src/parser/grammar.lalrpop
```diff
@@ -103,6 +103,8 @@ TyKind: SyntaxTyKind<'input> = {
 
     "string" <r:("(" <IntRange> ")")?> => SyntaxTyKind::Str(r),
     "buffer" <r:("(" <IntRange> ")")?> => SyntaxTyKind::Buf(r),
+    "vector" "(" <x_ty:Ty> "," <y_ty:Ty> ")" => SyntaxTyKind::Vector(Box::new(x_ty), Box::new(y_ty), Box::new(SyntaxTy { start: 0, end: 0, kind: SyntaxTyKind::Num(NumTy::U8, None)})),
+    "vector" "(" <x_ty:Ty> "," <y_ty:Ty> "," <z_ty:Ty> ")" => SyntaxTyKind::Vector(Box::new(x_ty), Box::new(y_ty), Box::new(z_ty)),
```
sasial-dev: You should be able to call `vector` without specifying types and it would default to `f32, f32, f32`

sasial-dev (changes requested):
Please fix clippy lints as well!

zap/src/parser/grammar.lalrpop
```diff
@@ -103,6 +103,13 @@ TyKind: SyntaxTyKind<'input> = {
 
     "string" <r:("(" <IntRange> ")")?> => SyntaxTyKind::Str(r),
     "buffer" <r:("(" <IntRange> ")")?> => SyntaxTyKind::Buf(r),
+    "vector" "(" ")" => SyntaxTyKind::Vector(
```
sasial-dev: For parity with other types such as `Instance` vs `Instance (BasePart)`, let's change it to:
```suggestion
    "vector" => SyntaxTyKind::Vector(
```

sasial-dev:
~~Also as vectors are natively testable please add an `irgen` test!~~ We can't do that until https://github.com/lune-org/lune/issues/299 but that doesn't need to block this PR.

docs/config/types.md
```diff
@@ -229,7 +229,16 @@ local CFrameSpecialCases = {
...
 ## Vectors
 
-Zap supports serializing Vector3s, although zap has both a Vector3 type and a Vector2 type. Zap does not allow the use of Vector2s, and instead only allows Vector3s as Vector2s don't use Luau's native vector type and instead are allocated on the heap. And thus, Zap's Vector2 type is almost the same as the Vector3 type, except it doesn't serialize the Z axis.
+Zap supports `vector`s with any numeric component types. The Z component is optional, and will result in a `0` if omitted.
+
+<CodeBlock code="type Position = vector(f64, f64, f64)" />
+<CodeBlock code="type Size = vector(u8, f64)" />
+
+Omitting all components will emit `vector(f32, f32, f32)`.
+<CodeBlock code="type Position = vector()" />
```
sasial-dev:
```suggestion
<CodeBlock code="type Position = vector" />
```

zap/src/irgen/mod.rs
```diff
@@ -477,6 +486,13 @@ impl Display for Expr {
 
 			Self::Color3(x, y, z) => write!(f, "Color3.fromRGB({}, {}, {})", x, y, z),
 			Self::Vector3(x, y, z) => write!(f, "Vector3.new({}, {}, {})", x, y, z),
+			Self::Vector(x, y, z) => write!(
+				f,
+				"vector.create({}, {}, {})",
+				x,
+				y,
+				z.clone().unwrap_or(Box::new(Expr::Num(0f64)))
```
sasial-dev:
```suggestion
				z.as_ref().unwrap_or(&Box::new(Expr::Num(0 as f64)))
```

zap/src/config.rs
```diff
@@ -291,6 +292,27 @@ impl<'src> Ty<'src> {
 			Self::Color3 => (12, Some(12)),
 			Self::Vector2 => (8, Some(8)),
 			Self::Vector3 => (12, Some(12)),
+			Self::Vector(x_ty, y_ty, z_ty) => {
+				let x_size = match **x_ty {
+					Ty::Num(numty, _) => numty.size(),
+					_ => unreachable!(),
```
sasial-dev: After testing locally, this `unreachable`/`panic` *does* trigger before the diagnostic is reported.
Maybe we should instead do:
```suggestion
					_ => 0,
```
Or maybe we default to the size of an F32
```suggestion
					_ => NumTy::F32.size(),
```

docs/config/types.md
```diff
@@ -229,7 +229,16 @@ local CFrameSpecialCases = {
...
+
+Zap also supports serializing Vector3s, although zap has both a Vector3 type and a Vector2 type. Zap does not allow the use of Vector2s, and instead only allows Vector3s as Vector2s don't use Luau's native vector type and instead are allocated on the heap. And thus, Zap's Vector2 type is almost the same as the Vector3 type, except it doesn't serialize the Z axis.
```
sasial-dev: It's a bit wordy and unclear, you can change it to something like:
```suggestion
Zap also supports serializing `Vector3`, and to not serialise the `Z` property of a `Vector3`, you can use the `Vector2` type.
```

sasial-dev (changes requested):

sasial-dev (approved):
Thank you!

PR #160, opened 2025-01-24
Update GitHub Actions

sasial-dev:
Some of the GitHub Actions we use will be depricated on the 30th of January, so we need to replace them.
Unfortunately, `actions/create-release` and `actions/upload-release-asset` are no longer maintained by GitHub so we had to switch to community ones.


jackhexed (approved):
lgtm

PR #161, opened 2025-01-25
Do not pass ID for unreliable events

sasial-dev:
Thanks to @nezuo for the optimisation idea!


zap/src/output/luau/server.rs
```diff
@@ -1086,21 +1049,28 @@ impl<'a> ServerOutput<'a> {
...
+			self.push_line(&format!("unreliable_{id} = Instance.new(\"UnreliableRemoteEvent\")"));
+			self.push_line(&format!(
+				"unreliable_{id}.Name = \"{}_UNRELIABLE_{id}\"",
+				self.config.remote_scope
+			));
+			self.push_line(&format!("unreliable_{id}.Parent = remotes"));
+			self.dedent();
+			self.push_line("end");
+			self.push("\n");
+		}
```
jackhexed: this is untenable due to the local limit

jackhexed (changes requested):

zap/src/output/luau/server.rs
sasial-dev: How would you suggest we get around this? Should we wrap it in a `do` `end` block?

jackhexed (approved):

PR #162, opened 2025-01-28
Disallow f64 vector components

Ketasaja:
This also adds a version requirement note to the heading of the vector docs. I think this should be replaced by a small box next to the heading.

### DIFF
--- a/zap/src/parser/convert.rs
+++ b/zap/src/parser/convert.rs
@@ -501,7 +501,7 @@ impl<'src> Converter<'src> {
 				match self.ty(x_ty) {
 					Ty::Num(numty, _) => {
 						if let NumTy::F64 = numty {
-							self.report(Report::AnalyzeInvalidVectorType {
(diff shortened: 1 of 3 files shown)


zap/src/parser/convert.rs
```diff
@@ -499,7 +499,16 @@ impl<'src> Converter<'src> {
...
 				match self.ty(x_ty) {
-					Ty::Num(_, _) => (),
+					Ty::Num(numty, _) => {
+						if let NumTy::F64 = numty {
+							self.report(Report::AnalyzeInvalidVectorType {
+								span: Span {
+									start: x_ty.start,
+									end: x_ty.end,
+								},
+							});
```
jackhexed: use a different report kind, this error will be very confusing so we should probably give some more specific explaination

jackhexed (approved):
I'll leave this for sasial to merge

sasial-dev (approved):
Thank you!

PR #139, opened 2024-11-23
Add tuples

nezuo:
TODO:
- [ ] TypeScript output
- [ ] Tooling output
- [ ] Code cleanup?
- [ ] Docs

PR #164, opened 2025-02-02
Add polling

Ketasaja:
Introduces a new option `call_default` and new `EvCall::Polling`.
```luau
for index, player, value in my_event.iter() do
end
```
### Can a table be directly returned to avoid iterator function call overhead?
Not really. Zap supports multiple parameters now.

### Why did you allow polling not to entirely drain the queue?
I talked to Marcus in ROSS because ECS and by extension Jecs is an important use case for polling. He indicated that the behavior he expected was that multiple polling locations would compete for events, which is what this PR implements. This behavior is also consistent with Blink's polling.

If there's an error handling a polling event, the user can poll again for further events, and won't receive stale events. This is important for ECS.

Avoiding an iterator function entirely used to be possible, but now that it isn't, so the complexity tradeoff may have changed.

I am not trying to be obstinate about this, if maintainers still don't like this design I'll drop it.

### Why return `index`?
I avoided some complexity from #139, where some sort of `non-nil` representation of `nil` would need to be returned to the user  on the client for events without data, or now for events with an optional first argument. Blink returns `index` too.


docs/config/events.md
```diff
@@ -1,23 +1,23 @@
...
 
 event TwoUnnamedParameters = {
-	from: Server,
-	type: Reliable,
-	call: ManyAsync,
-	data: (boolean, u32)
+ from: Server,
+ type: Reliable,
+ call: ManyAsync,
+ data: (boolean, u32)
```
sasial-dev: Out of scope for this PR, please check the diff of this file and revert unnecessary changes

docs/config/options.md
```diff
@@ -37,6 +37,15 @@ The paths are relative to the configuration file and should point to a lua(u) fi
...
 
+## `call_default` `[0.6.18+]`
+
+The default `call` field that will be used for events. See [call](events.html#call) for possible options.
+
+### Example
+
+<CodeBlock code = `opt call_default = "ManySync" />
+<CodeBlock code =`opt call_default = "Polling" />
+
```
sasial-dev, line 64: You might word this better than I can but

```suggestion
### Default

Requires an explicit call field defined on every event.
```

docs/config/options.md
```diff
@@ -45,15 +54,12 @@ This option changes the name of the remotes generated by Zap.
 
 <CodeBlock code = 'opt remote_scope = "ZAP"' />
 
-
```
sasial-dev: All these are also out of scope... please look at the diff for this file!

zap/src/output/luau/client.rs
```diff
@@ -241,10 +246,30 @@ impl<'src> ClientOutput<'src> {
...
+							"value".to_string()
+						} else {
+							format!("value{}", i)
+						}
+					})
+					.collect::<Vec<_>>();
+				if !ev.data.is_empty() {
+					for argument in arguments {
+						self.push_line(&format!(
+							"table.insert(polling_queues[{id}].events, if {} then {} else empty_value)",
```
sasial-dev: We know if data is `nil` ahead of time so we shouldn't need to do this?

zap/src/output/luau/client.rs
```diff
@@ -817,13 +863,144 @@ impl<'src> ClientOutput<'src> {
...
 		}
 	}
 
+	fn push_polling(&mut self) {
+		let filtered_evdecls = self
+			.config
+			.evdecls
+			.iter()
+			.filter(|evdecl| evdecl.from == EvSource::Server)
+			.filter(|evdecl| evdecl.call == EvCall::Polling);
```
sasial-dev:
```suggestion
			.filter(|evdecl| evdecl.from == EvSource::Server && evdecl.call == EvCall::Polling);
```

Ketasaja:
Indexing a circular buffer is `buffer[((index - 1) % size) + 1]` instead of `buffer[index % size]` because Luau arrays start at 1.

jackhexed (approved):

zap/src/output/luau/client.rs
```diff
@@ -18,6 +18,8 @@ struct ClientOutput<'src> {
 	var_occurrences: HashMap<String, usize>,
 }
 
+const INITIAL_POLLING_EVENT_CAPACITY: usize = 100;
```
sasial-dev: Why do we have an initial capacity?
Why 100?
We should move this to `src/output/luau/mod.rs`

zap/src/output/luau/client.rs
```diff
@@ -241,10 +245,91 @@ impl<'src> ClientOutput<'src> {
...
 		}
 
-		if ev.call == EvCall::SingleSync || ev.call == EvCall::SingleAsync {
-			self.push_line(&format!("if reliable_events[{id}] then"));
-		} else {
-			self.push_line(&format!("if reliable_events[{id}][1] then"));
+		match ev.call {
+			EvCall::SingleSync | EvCall::SingleAsync => self.push_line(&format!("if reliable_events[{id}] then")),
+			EvCall::ManySync | EvCall::ManyAsync => self.push_line(&format!("if reliable_events[{id}][1] then")),
+			EvCall::Polling => {
```
sasial-dev: There's a lot of logic under here that makes the file harder to follow, maybe we should extract everything inside thir block (L252-332) into a function?

Or maybe we extra all `EvCall::Polling` logic in `push_reliable_callback` into another function instead?

zap/src/output/luau/client.rs
```diff
@@ -817,13 +923,122 @@ impl<'src> ClientOutput<'src> {
...
+		let filtered_evdecls = self
+			.config
+			.evdecls
+			.iter()
+			.filter(|evdecl| evdecl.from == EvSource::Server)
+			.filter(|evdecl| evdecl.call == EvCall::Polling);
+
+		let is_polling_used = filtered_evdecls.clone().next().is_some();
+		if is_polling_used {
+			self.push_line("");
```
sasial-dev:
```suggestion
			self.push("\n");
```

zap/src/output/luau/client.rs
```diff
@@ -817,13 +923,122 @@ impl<'src> ClientOutput<'src> {
...
+	fn push_polling(&mut self) {
+		let filtered_evdecls = self
+			.config
+			.evdecls
+			.iter()
+			.filter(|evdecl| evdecl.from == EvSource::Server)
+			.filter(|evdecl| evdecl.call == EvCall::Polling);
+
+		let is_polling_used = filtered_evdecls.clone().next().is_some();
+		if is_polling_used {
```
sasial-dev: This is weird...

I'd either `.collect::<Vec<&EvDecl>>()` the iterator above and then use `filtered_evdecls.is_empty()` or use `let mut first: bool = false`

zap/src/output/luau/client.rs
```diff
@@ -817,13 +923,122 @@ impl<'src> ClientOutput<'src> {
...
+				return_names.push(format!(
+					"value_{}{}",
+					index + 1,
+					if let Some(name) = parameter.name {
+						format!("_{}", name)
+					} else {
+						String::from("")
+					}
+				));
+			}
```
sasial-dev, line 971: Can't we use the `get_named_values` function in `src/output/mod.rs` here?

zap/src/output/luau/client.rs
```diff
@@ -817,13 +923,122 @@ impl<'src> ClientOutput<'src> {
...
+			let arguments_size = return_names.len().max(1);
+
+			self.push_line(&format!("polling_queues[{id}] = {{"));
+			self.indent();
+
+			self.push_line(&format!(
+				"arguments = table.create({}),",
+				arguments_size * INITIAL_POLLING_EVENT_CAPACITY.max(1)
```
sasial-dev:
```suggestion
				(arguments_size * INITIAL_POLLING_EVENT_CAPACITY).max(1)
```

zap/src/output/luau/client.rs
```diff
@@ -817,13 +923,122 @@ impl<'src> ClientOutput<'src> {
...
+			));
+			self.push_line(&format!(
+				"queue_size = {},",
+				arguments_size * INITIAL_POLLING_EVENT_CAPACITY.max(1)
```
sasial-dev:
```suggestion
				(arguments_size * INITIAL_POLLING_EVENT_CAPACITY).max(1)
```

zap/src/output/luau/server.rs
```diff
@@ -250,10 +254,96 @@ impl<'a> ServerOutput<'a> {
...
 		}
 
-		if ev.call == EvCall::SingleSync || ev.call == EvCall::SingleAsync {
-			self.push_line(&format!("if reliable_events[{id}] then"))
-		} else {
-			self.push_line(&format!("for _, cb in reliable_events[{id}] do"))
+		match ev.call {
+			EvCall::SingleSync | EvCall::SingleAsync => self.push_line(&format!("if reliable_events[{id}] then")),
+			EvCall::ManySync | EvCall::ManyAsync => self.push_line(&format!("for _, cb in reliable_events[{id}] do")),
+			EvCall::Polling => {
```
sasial-dev: A lot of the same comments as `client.rs`, so please make sure they're applied in this file as well

zap/src/output/typescript/client.rs
```diff
@@ -99,25 +99,39 @@ impl<'src> ClientOutput<'src> {
...
 
 			self.push_line(&format!("export declare const {name}: {{", name = ev.name));
 			self.indent();
 
-			self.push_indent();
-			self.push(&format!("{set_callback}: ({callback}: ("));
+			if ev.call == EvCall::Polling {
+				self.push_indent();
+				let iter = self.config.casing.with("Iter", "iter", "iter");
+				self.push(&format!("{iter}: () => () => ["));
```
sasial-dev:
```suggestion
				self.push(&format!("{iter}: () => () => LuaTuple<["));
```

zap/src/output/typescript/client.rs
```diff
@@ -99,25 +99,39 @@ impl<'src> ClientOutput<'src> {
...
+				for (index, parameter) in ev.data.iter().enumerate() {
+					if index > 0 {
+						self.push(", ");
+					}
+					self.push_ty(&parameter.ty);
+				}
+				self.push("];\n");
```
sasial-dev:
```suggestion
				self.push("]>;\n");
```

zap/src/output/typescript/client.rs
```diff
@@ -99,25 +99,39 @@ impl<'src> ClientOutput<'src> {
...
+			} else {
+				let set_callback = match ev.call {
+					EvCall::SingleSync | EvCall::SingleAsync => {
+						self.config.casing.with("SetCallback", "setCallback", "set_callback")
+					}
+					EvCall::ManySync | EvCall::ManyAsync => self.config.casing.with("On", "on", "on"),
+					_ => todo!(),
```
sasial-dev:
```suggestion
					_ => unreachable!(),
```

zap/src/output/typescript/server.rs
```diff
@@ -174,24 +174,40 @@ impl<'a> ServerOutput<'a> {
...
 			self.indent();
 
-			let set_callback = match ev.call {
-				EvCall::SingleSync | EvCall::SingleAsync => {
-					self.config.casing.with("SetCallback", "setCallback", "set_callback")
+			if ev.call == EvCall::Polling {
+				self.push_indent();
+				let iter = self.config.casing.with("Iter", "iter", "iter");
+
+				self.push(&format!("{iter}: () => () => [Player, "));
```
sasial-dev:
```suggestion
				self.push(&format!("{iter}: () => () => LuaTuple<[Player, "));
```

zap/src/output/typescript/server.rs
```diff
@@ -174,24 +174,40 @@ impl<'a> ServerOutput<'a> {
...
+					if index > 0 {
+						self.push(", ");
+					}
+					self.push_ty(&parameter.ty);
 				}
-				EvCall::ManySync | EvCall::ManyAsync => self.config.casing.with("On", "on", "on"),
-			};
-			let callback = self.config.casing.with("Callback", "callback", "callback");
-			let player = self.config.casing.with("Player", "player", "player");
+				self.push("];\n");
```
sasial-dev:
```suggestion
				self.push("]>;\n");
```

zap/src/output/typescript/server.rs
```diff
@@ -174,24 +174,40 @@ impl<'a> ServerOutput<'a> {
...
+			} else {
+				let set_callback = match ev.call {
+					EvCall::SingleSync | EvCall::SingleAsync => {
+						self.config.casing.with("SetCallback", "setCallback", "set_callback")
+					}
+					EvCall::ManySync | EvCall::ManyAsync => self.config.casing.with("On", "on", "on"),
+					_ => todo!(),
```
sasial-dev:
```suggestion
					_ => unreachable!(),
```

zap/src/parser/convert.rs
```diff
@@ -233,6 +236,40 @@ impl<'src> Converter<'src> {
 		}
 	}
 
+	fn call_default_opt(&mut self, opts: &[SyntaxOpt<'src>]) -> Option<EvCall> {
+		if let Some(opt) = opts.iter().find(|opt| opt.name.name == "call_default") {
```
sasial-dev: nit: this should be a guard clause

zap/src/parser/convert.rs
```diff
@@ -364,7 +401,16 @@ impl<'src> Converter<'src> {
 		let name = evdecl.name.name;
 		let from = evdecl.from;
 		let evty = evdecl.evty;
-		let call = evdecl.call;
+		let call = if let Some(call) = evdecl.call {
+			call
+		} else if let Some(default) = self.call_default_opt(&self.config.opts.clone()) {
```
sasial-dev, line 421: We shouldn't need to `call_default_opt` again as it's at `config.call_default`?

zap/tests/files/polling.zap
```diff
@@ -0,0 +1,6 @@
+event polling = {
```
sasial-dev, line 1: Let's add another test case like this with a large number of events (see my other stress test) as I'm worried polling will hit the locals limit (which the lune test should check for)

sasial-dev (changes requested):
Sorry for the delay in getting back to you!
Thanks for all this work so far and apologies in advance for all the comments...

zap/tests/files/polling.zap
Ketasaja: @sasial-dev What part of the polling output are you concerned with the locals limit for when there's many events? The only thing I can think of is when an event returns tons of arguments, but that wouldn't depend on the number of events.

zap/src/output/luau/client.rs
Ketasaja: It's ambiguous whether there's an event without data or no event at all otherwise. On the server there is no ambiguity because there's always a player being returned.

zap/src/parser/convert.rs
Ketasaja: `config` doesn't exist yet, `self.config` is `SyntaxConfig`.

zap/tests/files/polling.zap
sasial-dev: Was concerned about 
```lua
			local queue = polling_queues[5]
			-- `arguments` is a circular buffer.
			-- `queue.arguments` can be replaced when it needs to grow.
			-- It's indexed like `arguments[((index - 1) % queue_size) + 1] because Luau has 1-based indexing.
			local arguments = queue.arguments
			local queue_size = queue.queue_size
			local read_cursor = queue.read_cursor
			local write_cursor = queue.write_cursor
			local unwrapped_write_end_cursor = write_cursor + 2
			local write_end_cursor = ((unwrapped_write_end_cursor - 1) % queue_size) + 1
			if (write_cursor < read_cursor and write_end_cursor >= read_cursor) or (unwrapped_write_end_cursor > queue_size and write_end_cursor >= read_cursor) then
				local new_queue_size = queue_size * 2
				local new_arguments = table.create(new_queue_size)
				local new_write_cursor
```
but it should be fine as it's inside an event callback
```lua
reliable.OnClientEvent:Connect(function(buff, inst)
	incoming_buff = buff
	incoming_inst = inst
	incoming_read = 0
	incoming_ipos = 0
	local len = buffer.len(buff)
	while incoming_read < len do
		local id = buffer.readu8(buff, read(1))
		if id == 0 then
```

cc @jackdotink

jackhexed: as long as it's not in the global scope it should be mostly fine - there might be edge cases but I'm not going to waste my time trying to find them, we'll fix them as they appear

sasial-dev:
Please fix merge conflicts!

zap/src/output/luau/client.rs
Ketasaja: Arbitrary default value that prevents moving tables a bunch as the queue grows. It probably does not matter either way.

zap/src/output/luau/client.rs
Ketasaja: That output would be less useful and could conflict with other variables unless a prefix were added later.

zap/src/output/luau/client.rs
Ketasaja: I genuinely don't understand how that code got there.

sasial-dev: Do you use an AI tool?

zap/src/output/luau/client.rs
sasial-dev: Maybe I'll switch it to use `get_unnamed_values` so the output is useful

sasial-dev: Change of plans, will leave for parity with the server :)

zap/src/output/luau/client.rs
```diff
@@ -817,13 +924,121 @@ impl<'src> ClientOutput<'src> {
...
+				if argument_index > 0 {
+					self.push(", ");
+				}
+				self.push(&format!(
+					"arguments[{}]",
+					if argument_index > 0 {
+						format!("((read_cursor + {} - 1) % queue_size) + 1", argument_index)
+					} else {
+						String::from("read_cursor")
+					}
```
sasial-dev: Some concerns:

1. isn't reading from `read_cursor` reading the value from before where it should be (with the `+1`)?
2. Let's add a fast call (`else if`) if `i == 1` so we don't get `queue.read_cursor = ((queue.read_cursor + 1 - 1) % queue.queue_size) + 1`
3. The `+ {} - 1` can be evaluated ahead of time on the rust side

sasial-dev (changes requested):
Rust code looks good, just got some concerns with the generated code.
This single comment appiles to add reads/writes of the polling queue on client/server

sasial-dev (approved):
Thank you so much for all your work and patience with this!
I think I'm happy enough for it to be merged.

PR #165, opened 2025-03-01
Fix tagged variant index serialization

daimond113:
The index for tagged variants was hardcoded to u8 in serialization, despite supporting other NumTys in the deserialization code.


sasial-dev (approved):
Thank you!

PR #166, opened 2025-03-01
Optimize length storage

daimond113:
Storing lengths was hardcoded to u16, even when a different type would be better. So, now if the range of length is known it'll be used to serialize the length in the fitting size.

This currently requires explicit minimum values (`0..255`). Should a minimum of 0 be default?

### DIFF
--- a/zap/src/config.rs
+++ b/zap/src/config.rs
@@ -428,7 +428,7 @@ impl Range {
 	}
 
 	pub fn numty(&self) -> Option<NumTy> {
-		Some(NumTy::from_f64(self.min.unwrap_or_default(), self.max?))
+		Some(NumTy::from_f64(self.min.unwrap_or(0.0), self.max?))
 	}
 }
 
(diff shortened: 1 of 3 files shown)


sasial-dev:
I think that a minimum of 0 should be the default, yes. As you'll never have a negative length string/buffer/array etc.

zap/src/config.rs
```diff
@@ -426,6 +426,10 @@ impl Range {
 			None
 		}
 	}
+
+	pub fn numty(&self) -> Option<NumTy> {
+		Some(NumTy::from_f64(self.min.unwrap_or_default(), self.max?))
```
sasial-dev: I know they are equivalent, but to make it more clear for future readers:
```suggestion
		Some(NumTy::from_f64(self.min.unwrap_or(0.0), self.max?))
```

sasial-dev (approved):
Other than this comment, LGTM

sasial-dev:
Thank you!

PR #168, opened 2025-03-01
Fix tooling deserializer for unreliable events

nezuo:
The check for whether the event is received on the client or server was opposite of what it should be.

I also added an error message if the user passes a non-zap remote instance to the tooling deserializer.

### DIFF
--- a/zap/src/output/tooling.rs
+++ b/zap/src/output/tooling.rs
@@ -462,11 +462,6 @@ impl<'src> ToolingOutput<'src> {
 
 		self.push_unreliable_events();
 
-		self.push_line("else");
-		self.indent();
-		self.push_line("error(`Unknown remote instance: {remote_instance}`)");
-		self.dedent();
-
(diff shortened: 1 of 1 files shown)


zap/src/output/tooling.rs
```diff
@@ -462,6 +462,11 @@ impl<'src> ToolingOutput<'src> {
 
 		self.push_unreliable_events();
 
+		self.push_line("else");
+		self.indent();
+		self.push_line("error(`Unknown remote instance: {remote_instance}`)");
+		self.dedent();
+
```
sasial-dev: Don't we already check for remote instance at L391-394 of `tooling.rs`?

nezuo: Yeah, you're right. The error was only relevant because of the bug, now it's not.

sasial-dev (approved):
Thank you!

PR #169, opened 2025-03-02
Support quoted identifiers with spaces

daimond113:
Adds support to quoting identifiers which are:
- keys of structs
- keys of tagged enums
- values of unit enums

Quoted identifiers are useful because they're allowed to contain spaces.

Additionally, fixes the types for tags of enums containing spaces.


jackhexed (approved):

sasial-dev (changes requested):
Please add tests

sasial-dev (approved):
LGTM!

sasial-dev:
Thank you!

PR #170, opened 2025-03-02
Fix TypeScript types being emitted without typescript option

daimond113:
The `types_output` option always emitted TypeScript definitions, even if the `typescript` option was disabled.


sasial-dev (approved):
Thank you!
That's an embarrasing oversight...

PR #171, opened 2025-03-12
Fix unit enum size

daimond113:
Subtracts one from the variants' length, to correct the size not being accurate to what is used everywhere else.


sasial-dev (approved):
Thank you!

PR #172, opened 2025-03-12
Optimize table-like length storage

daimond113:
Optimizes length storage for table-like types (map, set). It works by getting the amount of possible keys (number of variants for enums) to use for the length storage.

### DIFF
--- a/zap/src/config.rs
+++ b/zap/src/config.rs
@@ -33,7 +33,7 @@ pub struct Config<'src> {
 	pub disable_fire_all: bool,
 }
 
-impl Config<'_> {
+impl<'src> Config<'src> {
 	pub fn server_reliable_count(&self) -> usize {
 		let reliable_count = self
 			.evdecls
@@ -74,6 +74,17 @@ impl Config<'_> {
 
 	pub fn client_reliable_ty(&self) -> NumTy {
 		NumTy::from_f64(0.0, self.client_reliable_count() as f64 - 1.0)
+	}
+
+	pub fn resolve_ty<'a>(&'a self, ty: &'a Ty<'src>) -> &'a Ty<'src> {
+		match ty {
+			Ty::Ref(name) => self
+				.tydecls
+				.iter()
+				.find(|decl| decl.name == *name)
+				.map_or(ty, |decl| &decl.ty),
+			_ => ty,
+		}
 	}
 }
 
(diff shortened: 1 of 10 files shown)


zap/src/config.rs
```diff
@@ -33,7 +33,7 @@ pub struct Config<'src> {
 	pub disable_fire_all: bool,
 }
 
-impl Config<'_> {
+impl<'src> Config<'src> {
```
sasial-dev: Why did we remove the static lifetime?

zap/src/config.rs
```diff
@@ -313,6 +324,24 @@ impl<'src> Ty<'src> {
...
+				NumTy::U8 => Some(NumTy::U8),
+				NumTy::U16 => Some(NumTy::U16),
+				NumTy::U32 => None,
+				NumTy::I8 => Some(NumTy::U8),
+				NumTy::I16 => Some(NumTy::U16),
+				NumTy::I32 => None,
+			},
+			_ => None,
+		}
+	}
```
sasial-dev: This function is really weird.

zap/src/irgen/des.rs
```diff
@@ -7,6 +7,7 @@ struct Des<'src> {
 	checks: bool,
 	buf: Vec<Stmt>,
 	var_occurrences: &'src mut HashMap<String, usize>,
+	config: &'src Config<'src>,
```
sasial-dev: Let's not pass the whole config into `irgen`

zap/src/irgen/ser.rs
```diff
@@ -206,7 +212,7 @@ impl Ser<'_> {
 				self.push_stmt(Stmt::End);
 
 				self.push_stmt(Stmt::Call(
-					Var::from("buffer").nindex("writeu16"),
+					Var::from("buffer").nindex(format!("writeu{}", length_size * 8)),
```
sasial-dev: `NumTy` itself has a method to display the correct `write` thing

sasial-dev (changes requested):
Thank you for the idea - but I think we need to reconsider how it's actually executed.

I'm not a big fan of `resolve_ty`/ `variants_size` personally - resolving `tydecls` based upon going back to the config so many times over feels stupid! I also don't want to pass the whole config to `irgen`.

One idea I had - maybe we should consider adding a `length` argument to the `ty`? Then we can assign it in `convert.rs` or whatever.
Otherwise, let me know if you have any other ideas on how we should approach this.

I know this would sorta backpedal and require a refactor of your work on #166, but I think it's for the best.

daimond113:
I removed the number optimization since the formula for the length of a map is actually `variants_size + 1` (because length 0 needs to be stored as well).

sasial-dev:
I wonder if we can (or should!) make our optimisations more aggressive?

e.g. length should only be stored in a `u8` for:
```
event Test = {
    from: Client,
    type: Reliable,
    call: SingleSync,
    data: u8[255..270]
}
```

daimond113:
I've been looking into making the NumTy algorithm smarter like that, but I think it's better if it becomes a separate PR.

sasial-dev (approved):
Thank you!
This lays the groundwork nicely for further optimisations.

PR #173, opened 2025-03-15
Fix inaccurate Ty::size values after #166

daimond113:
Fixes the length size changes from #166 not being factored in in the Ty::size method.


sasial-dev (approved):
Thank you!

PR #174, opened 2025-03-17
Fix buffer length check with `write_checks` on

Ketasaja:
```
event CustomTransform = {
  from: Client,
  type: Unreliable,
  call: ManyAsync,
  data: struct {
    Transform: buffer(19)
  }
}
```


sasial-dev (approved):
Thank you!

PR #176, opened 2025-03-17
Fix unreliable polling

Ketasaja:
Fixes polling queue indexing and unreliable polling.

### DIFF
--- a/zap/src/config.rs
+++ b/zap/src/config.rs
@@ -156,15 +156,6 @@ pub enum EvType {
 	Unreliable,
 }
 
-impl EvType {
-	pub fn name(&self) -> &str {
-		match &self {
-			EvType::Reliable => "reliable_event",
-			EvType::Unreliable => "unreliable_event",
-		}
-	}
-}
-
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum EvCall {
 	SingleSync,
--- a/zap/src/config.rs
+++ b/zap/src/config.rs
@@ -156,6 +156,15 @@ pub enum EvType {
 	Unreliable,
 }
 
+impl EvType {
+	pub fn name(&self) -> &str {
+		match &self {
+			EvType::Reliable => "reliable",
+			EvType::Unreliable => "unreliable",
+		}
+	}
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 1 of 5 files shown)


zap/src/config.rs
```diff
@@ -156,6 +156,15 @@ pub enum EvType {
...
 
+impl EvType {
+	pub fn name(&self) -> &str {
+		match &self {
+			EvType::Reliable => "reliable",
+			EvType::Unreliable => "unreliable",
+		}
+	}
+}
+
```
sasial-dev: We should instead match `zap/src/output/luau/mod.rs` at L218-L231

sasial-dev (changes requested):

zap/src/config.rs
Ketasaja: I kept it singular because that makes more sense.

sasial-dev: I'm not sure you got what I meant - sorry about being unclear.
I'm happy with `reliable_polling_queue`or whatever it was. But, instead of adding a method to `EvType`, we should do what was done in `zap/src/output/luau/mod.rs` at L218-L231:

```rs
fn events_table_name<'a>(evdecl: &EvDecl) -> &'a str {
	match evdecl.evty {
		EvType::Reliable => "reliable_events",
		EvType::Unreliable => "unreliable_events",
	}
}

fn event_queue_table_name<'a>(evdecl: &EvDecl) -> &'a str {
	match evdecl.evty {
		EvType::Reliable => "reliable_event_queue",
		EvType::Unreliable => "unreliable_event_queue",
	}
}
```

zap/src/output/luau/client.rs
```diff
@@ -563,68 +564,72 @@ impl<'src> ClientOutput<'src> {
 			self.push_stmts(statements);
 		}
 
-		if ev.call == EvCall::SingleSync || ev.call == EvCall::SingleAsync {
-			self.push_line(&format!("if unreliable_events[{id}] then"));
+		if ev.call == EvCall::Polling {
+			self.push_polling_event(ev)
 		} else {
```
sasial-dev, line 569: Why is this different to how reliable polling events are pushed?
Shouldn't unreliable/reliable events should have (roughly) the same codepath for listeners?

sasial-dev (approved):
Thank you!

PR #178, opened 2025-03-22
Add OR type syntax

### LINKED ISSUE #46: [FEAT] Addition of an `OR` operator for types
## Describe your feature
The addition of an `OR` operator for types will create better data definitions with less overhead, compared to tagged enum alternatives. The operator can function similar to Luau i.e. `type bool_or_u8 = boolean | u8`

## Alternatives
Using tagged enums will in essence function as an `OR`, however there is more overhead. 
i.e.
```lua
enum "Type" {
    Number {
        Value: f64,
    },

    String {
        Value: string,
    },
}
```
WILL CREATE -->
```lua
({
	Type: "Number",
	Value: (number),
} | {
	Type: "String",
	Value: (string),
})
```
meaning that there is now an overhead of both a string and the value's type.

daimond113:
Adds an `OR` type for types. Fixes #46.


sasial-dev:
Just be warned this now has merge conflicts with #172.

zap/src/irgen/des.rs
```diff
@@ -291,6 +291,28 @@ impl Des<'_> {
 				self.push_struct(struct_ty, into)
 			}
 
+			Ty::Or(or_tys) => {
+				self.push_assign(into.clone(), Expr::EmptyTable);
```
sasial-dev: Why? We shouldn't need to assign `value` at the start.

zap/src/irgen/ser.rs
```diff
@@ -278,6 +278,68 @@ impl Ser<'_> {
...
+									Some("IsA".to_string()),
+									vec![Expr::Str(class.to_string())],
+								))
+							}
+							cond
+						}
+						PrimitiveTy::Unit(variants) => variants
+							.iter()
+							.map(|variant| from_expr.clone().eq(Expr::Str(variant.to_string())))
+							.reduce(|acc, ex| acc.or(ex))
```
sasial-dev: This adds overhead and makes the generated emit weird.
```lua
			if Value == "a" or Value == "v" or Value == "d" then
				alloc(1)
				buffer.writeu8(outgoing_buff, outgoing_apos, 0)
				if Value == "a" then
					alloc(1)
					buffer.writeu8(outgoing_buff, outgoing_apos, 0)
				elseif Value == "v" then
					alloc(1)
					buffer.writeu8(outgoing_buff, outgoing_apos, 1)
				elseif Value == "d" then
					alloc(1)
					buffer.writeu8(outgoing_buff, outgoing_apos, 2)
				else
					error("Invalid enumerator")
				end
```
To save bandwidth and performance, the `.reduce` shoudl be removed and all of those mapped variants should be appended. I think you can do it easily if `condition` is a `Vec`.

zap/src/parser/convert.rs
```diff
@@ -682,12 +675,10 @@ impl<'src> Converter<'src> {
 								name,
 							});
 
-							return Ty::Unknown;
+							return Ty::Ref(name, Box::new(Ty::Opt(Box::new(Ty::Unknown))));
```
sasial-dev, line 682: I don't disagree with this change - just curious to know why you did this?

zap/src/config.rs
```diff
@@ -324,10 +347,34 @@ impl<'src> Ty<'src> {
...
+			Ty::DateTime => PrimitiveTy::Name("DateTime"),
+			Ty::Boolean => PrimitiveTy::Name("boolean"),
+			Ty::Color3 => PrimitiveTy::Name("Color3"),
+			Ty::Vector2 => PrimitiveTy::Name("Vector2"),
+			Ty::Vector3 => PrimitiveTy::Name("Vector3"),
+			Ty::Vector(..) => PrimitiveTy::Name("Vector3"),
+			Ty::AlignedCFrame => PrimitiveTy::Name("CFrame"),
+			Ty::CFrame => PrimitiveTy::Name("CFrame"),
+			Ty::Instance(class) => PrimitiveTy::Instance(*class),
+			Ty::Enum(Enum::Unit(variants)) => PrimitiveTy::Unit(variants.clone()),
```
sasial-dev: Tagged enums should be primative? As we can differentiate between them at runtime?

zap/src/parser/reports.rs
```diff
@@ -338,6 +353,17 @@ impl Report<'_> {
...
+
+			Self::AnalyzeOrDuplicateType { prev_span, dup_span } => {
+				vec![
+					Label::secondary((), prev_span.clone()).with_message("previous type usage"),
+					Label::primary((), dup_span.clone()).with_message("duplicate type usage"),
+				]
+			}
+
+			Self::AnalyzeOrNonPrimitiveType { span } => {
+				vec![Label::primary((), span.clone()).with_message("non-primitive type")]
```
sasial-dev: This is unclear to me. I think we might just need to explain it better in the docs maybe.

sasial-dev (changes requested):

zap/src/config.rs
daimond113: They're checked by their variants, and if a variant overlaps a duplicate type error is reported.

zap/src/parser/reports.rs
daimond113: I don't think there's much I can do here to explain it better. A help note might improve it, but explaining it in the docs will probably be a better choice.

zap/src/parser/convert.rs
daimond113: Leftovers from a merge conflict, but I think it's better since it is more specific.

zap/src/irgen/des.rs
daimond113: Copied over from the cases for enums and structs above.

zap/src/config.rs
```diff
@@ -261,27 +270,50 @@ impl<'src> Ty<'src> {
...
+				for ty in or_tys {
+					let (ty_min, ty_max) = ty.size(recursed);
+
+					if ty_min < min {
+						min = ty_min;
+					}
+
+					if let Some(ty_max) = ty_max {
+						if let Some(current_max) = max {
+							max = Some(ty_max + current_max);
```
sasial-dev: Doesn't this needs to check if what max is being set to is larger than the current value of max?

zap/src/irgen/des.rs
```diff
@@ -291,6 +291,75 @@ impl Des<'_> {
...
+								self.push_assign(
+									into.clone().eindex(Expr::Str(tag.to_string())),
+									Expr::Str(variant.to_string()),
+								);
+								self.push_struct(&data, into.clone());
+							}
+
+							continue;
+						}
+						_ => {}
```
sasial-dev: Instead of shadowing on line 302, maybe we do `i_offset += 1` here, and remove the `enumerate` on line 301.

zap/src/irgen/ser.rs
```diff
@@ -278,6 +278,105 @@ impl Ser<'_> {
...
+						vec![Expr::from(from.clone())],
+					)),
+				);
+
+				let mut unknown_i = None;
+				let mut initial_if = true;
+				let mut i_offset = 0usize;
+
+				for (i, ty) in or_tys.iter().enumerate() {
+					let i = i + i_offset;
```
sasial-dev: Same deal here.

zap/src/parser/convert.rs
```diff
@@ -697,6 +688,89 @@ impl<'src> Converter<'src> {
...
+					(PrimitiveTy::Unknown, _) => Ordering::Greater,
+					(_, PrimitiveTy::Unknown) => Ordering::Less,
+					(PrimitiveTy::Instance(None), _) => Ordering::Greater,
+					(_, PrimitiveTy::Instance(None)) => Ordering::Less,
+					(PrimitiveTy::Enum(Enum::Unit(..)), _) => Ordering::Less,
+					(_, PrimitiveTy::Enum(Enum::Unit(..))) => Ordering::Greater,
+					(PrimitiveTy::Enum(Enum::Tagged { .. }), _) => Ordering::Less,
+					(_, PrimitiveTy::Enum(Enum::Tagged { .. })) => Ordering::Greater,
+					_ => Ordering::Equal,
+				});
```
sasial-dev: Does this actually guarantee that order? 
All it says it that unknown is at the end and anything before unknown is before etc.
It might make more sense to just group the `PrimitiveTy` into vecs maybe?

The other reason I'd suggest to consider as well is because if an `unknown` is present inside an opt or `?` is present after an opt, we should just have extra cases inside the `or` so we don't need to use two `u8`s.

Emit-wise this is currently the case of an unknown:
```lua
				else
					alloc(1)
					buffer.writeu8(outgoing_buff, outgoing_apos, 10)
					if Value == nil then
						alloc(1)
						buffer.writeu8(outgoing_buff, outgoing_apos, 0)
					else
						alloc(1)
						buffer.writeu8(outgoing_buff, outgoing_apos, 1)
						table.insert(outgoing_inst, Value)
					end
				end
```

And in the case of a `?` after the `OR` block:
```lua
			if Value == nil then
				alloc(1)
				buffer.writeu8(outgoing_buff, outgoing_apos, 0)
			else
				alloc(1)
				buffer.writeu8(outgoing_buff, outgoing_apos, 1)
				local ty_name_1 = typeof(Value)
```

zap/tests/irgen/mod.rs
```diff
@@ -354,3 +354,35 @@ async fn test_tagged_enum_quoted() {
 
 	runtime.run("Zap", output).await.unwrap();
 }
+
```
sasial-dev, line 359: Can we add some more complex tests like what you've been testing with locally?
As an example:
```
event Test = {
    from: Client,
    type: Reliable,
    call: SingleSync,
    data: (string | unknown | buffer | u16 | struct { bar: u8 } | enum { foo, bar, baz } | Instance (Part) | enum "test" {
        a { 
            b: u8
        },
        c { 
            d: u8
        }
    })?
}
```

sasial-dev (changes requested):
The code is looking really good so far:

sasial-dev (approved):
Thank you!

PR #179, opened 2025-03-23
Fix string length constraint example in documentation

AzzaDeveloper:
Originally reported by nogo (nogo. on Discord)

### DIFF
--- a/docs/config/types.md
+++ b/docs/config/types.md
@@ -98,9 +98,9 @@ Strings are defined using the word `string`. For example:
 
 <CodeBlock code="string" />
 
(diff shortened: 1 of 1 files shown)


sasial-dev (commented):
I wonder if it's better to remove the `(like a username)` bit instead?
I'm curious to know your thoughts.

AzzaDeveloper:
> I wonder if it's better to remove the `(like a username)` bit instead? I'm curious to know your thoughts.

I looked up to see whether there were constraints regarding username character length and while I couldn't find any official resources that gave a specific number, a lot of user generated posts point to the 20 character limit (which DisplayName shares), so I've elected to keep the `(like a username)` bit to be consistent with what I assume is common belief for users :D

sasial-dev (approved):
Thank you!

PR #180, opened 2025-03-29
Add offsets to NumTy algorithm

daimond113:
This PR makes the NumTy algorithm smarter by storing only the offset from the minimum value of the range. This means that `[255..265]` will now store a `u8` of `[0, 10]` and 255 will be added or subtracted as needed.

### DIFF
--- a/zap/src/config.rs
+++ b/zap/src/config.rs
@@ -217,7 +217,7 @@ impl<'src> Ty<'src> {
 				if let Some(exact) = len.exact() {
 					(exact as usize, Some(exact as usize))
 				} else {
-					let len_numty = len.numty().map(|numty| numty.numty).unwrap_or(NumTy::U16);
+					let len_numty = len.numty().map(|(numty, ..)| numty).unwrap_or(NumTy::U16);
 
 					(
 						len.min().map(|_| len_numty.min() as usize).unwrap_or(0) + len_numty.size(),
@@ -230,7 +230,7 @@ impl<'src> Ty<'src> {
 				if let Some(exact) = len.exact() {
 					(exact as usize, Some(exact as usize))
 				} else {
-					let len_numty = len.numty().map(|numty| numty.numty).unwrap_or(NumTy::U16);
+					let len_numty = len.numty().map(|(numty, ..)| numty).unwrap_or(NumTy::U16);
 
 					(
 						len.min().map(|_| len_numty.min() as usize).unwrap_or(0) + len_numty.size(),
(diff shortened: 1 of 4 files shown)


zap/src/config.rs
```diff
@@ -482,7 +487,13 @@ impl NumTy {
 			NumTy::U32
 		} else {
 			NumTy::F64
-		}
+		};
+
+		(numty, offset)
```
sasial-dev: It's weird to return a tuple from `from_f64` and then have to invent a new method `from_f64_max` to *not* do that.

Consider either:
1. NumTy becomes a `struct` with an optional range and then the type of number
2. NumTy stays as an enum, but inside the args of each item it has an optional range

sasial-dev (changes requested):

zap/src/config.rs
```diff
@@ -230,29 +230,29 @@ impl<'src> Ty<'src> {
...
 			}
 
 			Self::Arr(ty, len) => {
 				let (ty_min, ty_max) = ty.size(tydecls, recursed);
 				let len_min = len.min().map(|min| min as usize).unwrap_or(0);
-				let len_size = len.numty().unwrap_or(NumTy::U16).size();
+				let len_numty = len.numty().map(|(numty, ..)| numty).unwrap_or(NumTy::U16);
 
 				if let Some(exact) = len.exact() {
 					(ty_min * (exact as usize), ty_max.map(|max| ty_max.unwrap() * max))
```
sasial-dev: I don't think it's your fault, but I think this is wrong as it's just doing `ty_max * ty_max`

zap/src/config.rs
```diff
@@ -230,29 +230,29 @@ impl<'src> Ty<'src> {
...
-				} else if let Some(len_max) = len.max() {
+				} else {
 					(
-						ty_min * len_min + len_size,
-						ty_max.map(|ty_max| ty_max * (len_max as usize) + len_size),
+						ty_min * len_min + len_numty.size(),
+						ty_max
+							.and_then(|ty_max| len.max().map(|max| (ty_max, max as usize)))
+							.map(|(ty_max, max)| ty_max * max + len_numty.size()),
```
sasial-dev: A little nit - we can simplify the logic a little here so we don't have to map twice:
```suggestion
						ty_max
							.zip(len.max())
							.map(|(ty_max, max)| ty_max * max as usize + len_numty.size()),
```

zap/src/config.rs
```diff
@@ -432,8 +432,17 @@ impl Range {
 		}
 	}
 
-	pub fn numty(&self) -> Option<NumTy> {
-		Some(NumTy::from_f64(self.min.unwrap_or(0.0), self.max?))
+	pub fn numty(&self) -> Option<(NumTy, f64)> {
+		let min = self.min.unwrap_or(0.0);
+		let max = self.max?;
```
sasial-dev: When `self.max` isn't defined, why don't we return `(NumTy::U16, 0)` here instead of this function returning a `None`?

sasial-dev (changes requested):

sasial-dev (approved):
Thank you!

PR #181, opened 2025-04-07
Fix unwrap on None variant sizes

daimond113:
Fixes a mistake from #172.


sasial-dev (approved):
I totally didn’t forget to write a review comment for the unwrap…
Thank you for the pickup!

PR #182, opened 2025-04-11
Add ordered unreliables

### LINKED ISSUE #61: [FEAT] Add "OrderedUnreliable" Type
## Describe your feature
An ordered, unreliable event type would help developers handle data that immediately expires. It would reduce boiler-plate code in which expired data packets are thrown out and is a natural addition to a networking generated-library. Consider the following example demonstrating the problem such an addition would solve:
- Some client sends three unordered packets modeling an object's physics data at a given moment in time via an unreliable remote in the following order: packet 1, packet 2, packet 3. Note that data in packet 1 is thus older than data in packet 2, and likewise to packets 2 and 3.
- The server receives the packets out of order (say 2, 1, 3), however, nonetheless processes the data (in the order received) and replicates it to other clients.
- Other clients apply the physics data in the order received (say 2, 1, 3 again). _Because packet 2's data is older than packet 1's, some "rubberbanding" occurs when applying the physics data to the game world._

## Alternatives
1. The control alternative would be to not have this type and simply leave packet ordering to developers. This is fine, but as said before such a type would be a natural addition to a networking library and would reduce boiler-plate code.
2. A new field ("order") could be added to the event constructor. I probably wouldn't do this because such a field would only be relevant for events with _type_ set to Unreliable.

### Approach 1
A single, u32 "packetCount" variable for each event, incremented by the sender with each fire and used by the receiver to determine what order packets were sent, throwing out packets with a packetCount less than what has already been read.

### Approach 2 (recommended)
Same functionality as before, but packetCount would be u16 and reset to 0 after the 65,535 limit is reached. When the receiver is at or near the limit, it would accept the first packet with a substantially small packetCount value (0, 1, 2, etc., not just 0 because the packet with packetCount 0 could fail to deliver).  The receiver would not accept packets with huge packetCount differences (in case packet A has a packetCount of 65,535 and packet B's is 0, and packet B is delivered before packet A). At max sending speeds (every heartbeat), this result would result in a reset every ~18 min. I'm recommending this approach because it seems to be the sweet spot for timing resets and avoiding additional overhead.

### Approach 3
Same as before, except a u8 packetCount is used. This would mean packetCount would reset every ~4 seconds at max speeds.

## Additional context
Feel free to message me on discord: I'm @reydelcodigo.

daimond113:
Adds ordered unreliables, which work by serializing an order id (u16) that gets reset whenever it overflows the maximum size. Packets received with older and/or equal order ids, and without a significant drop (a reset from the maximum value to 0) are not handled.

Fixes #61.


zap/src/output/luau/client.rs
```diff
@@ -716,9 +733,35 @@ impl<'src> ClientOutput<'src> {
 		self.push_line(&format!("buffer.write{}(outgoing_buff, outgoing_apos, {id})", num_ty));
 	}
 
+	fn push_get_order_id(&mut self, id: usize) {
+		self.push_line(&format!("local order_id = outgoing_ids[{id}]"));
+		self.push_line(&format!(
+			"if order_id and order_id <= {} then",
```
sasial-dev: Consider doing what we do with function call IDs where instead of branching we use the mod (`%=`) operator.

zap/src/output/luau/client.rs
```diff
@@ -716,9 +733,35 @@ impl<'src> ClientOutput<'src> {
...
+		));
+	}
+
 	fn push_write_evdecl_event_id(&mut self, ev: &EvDecl) {
-		if ev.evty == EvType::Reliable {
-			self.push_write_event_id(ev.id);
+		match ev.evty {
+			EvType::Reliable => self.push_write_event_id(ev.id),
+			EvType::Unreliable(true) => self.push_get_order_id(ev.id),
+			_ => {}
```
sasial-dev: In case we add more to `EvType` in the future
```suggestion
			EvType::Unreliable(false) => {}
```

zap/src/output/luau/client.rs
```diff
@@ -758,13 +801,16 @@ impl<'src> ClientOutput<'src> {
...
+		if let EvType::Unreliable(ordered) = ev.evty {
 			self.push_line("local saved = save()");
 			self.push_line("load_empty()");
+			if ordered {
+				self.push_write_order_id();
+			}
 		}
 
-		self.push_write_evdecl_event_id(ev);
-
```
sasial-dev: You're loading an empty buffer *after* writing the event ID

zap/src/output/luau/server.rs
```diff
@@ -532,6 +535,27 @@ impl<'a> ServerOutput<'a> {
...
+				"if {last} and order_id <= {last} and {last} - order_id < {} then",
+				(UNRELIABLE_ORDER_NUMTY.max() / 2.0).floor(),
+			));
+			self.indent();
+			self.push_line("return");
+			self.dedent();
+			self.push_line("end");
+			self.push_line(&format!("{last} = order_id"));
+			self.push_line("player_map[player] = save()");
+			self.push_line("load(saved)");
```
sasial-dev, line 556: Calling save twice is hacky!
If you don't want to increment when calling `read`, just use the `incoming_read` variable.

zap/src/output/luau/server.rs
```diff
@@ -615,9 +639,18 @@ impl<'a> ServerOutput<'a> {
...
+		));
+	}
+
 	fn push_write_evdecl_event_id(&mut self, ev: &EvDecl) {
-		if ev.evty == EvType::Reliable {
-			self.push_write_event_id(ev.id);
+		match ev.evty {
+			EvType::Reliable => self.push_write_event_id(ev.id),
+			EvType::Unreliable(true) => self.push_alloc_order_id(),
+			_ => {}
```
sasial-dev:
```suggestion
			EvType::Unreliable(false) => {}
```

zap/src/output/luau/server.rs
```diff
@@ -643,6 +676,29 @@ impl<'a> ServerOutput<'a> {
...
+		self.push_line(&format!(
+			"if order_id and order_id <= {} then",
+			UNRELIABLE_ORDER_NUMTY.max()
+		));
+		self.indent();
+		self.push_line(&format!("outgoing_ids[{id}] += 1"));
+		self.dedent();
+		self.push_line("else");
+		self.indent();
+		self.push_line(&format!("outgoing_ids[{id}] = 1"));
```
sasial-dev: Same deal here, use the `%=` operator instead of branching

sasial-dev (changes requested):
My general piece of feedback in this PR is to not do these hacky `save` and `load` calls.

zap/src/output/luau/base.luau
```diff
@@ -79,18 +83,24 @@ local function load(data: {
...
+	incoming_ids = data.incoming_ids
 end
 
 local function load_empty()
 	outgoing_buff = buffer.create(64)
 	outgoing_used = 0
 	outgoing_size = 64
 	outgoing_inst = {}
+	outgoing_ids = {}
+	incoming_ids = {}
```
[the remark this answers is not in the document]
daimond113: `table.create`ing a holey array is wasteful, since it'll use the hashmap portion of the table for non-continuous indices.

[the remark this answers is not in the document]
sasial-dev: Thanks for the pickup!

zap/src/output/tooling.rs
```diff
@@ -545,7 +545,7 @@ impl<'src> ToolingOutput<'src> {
 			.config
 			.evdecls
 			.iter()
-			.filter(|ev_decl| ev_decl.evty == EvType::Unreliable)
+			.filter(|ev_decl| matches!(ev_decl.evty, EvType::Unreliable(_)))
```
sasial-dev, line 566: The tooling file will need to deal with ordered event ID (see how we deal with `call_id`)

zap/src/output/luau/client.rs
```diff
@@ -758,9 +775,23 @@ impl<'src> ClientOutput<'src> {
 		self.push(")\n");
 		self.indent();
 
-		if ev.evty == EvType::Unreliable {
+		if let EvType::Unreliable(ordered) = ev.evty {
+			let id = ev.id;
+			if ordered {
+				self.push_line(&format!(
+					"local order_id = ((outgoing_ids[{id}] or -1) + 1) % {}",
+					UNRELIABLE_ORDER_NUMTY.max() + 1.0
```
sasial-dev, line 783: You shouldn't need to `+1` here?

sasial-dev (changes requested):

sasial-dev (approved):
Thank you!

PR #184, opened 2025-04-12
Add messages to asserts

daimond113:
Adds & requires that asserts are given a message to ensure Selene tests pass and errors are easier to debug.


sasial-dev (approved):
Thank you!

PR #185, opened 2025-04-13
Improve map & set size accuracy

daimond113:
Gives a hint of the maximum sizes of sets and maps with the previous `variants_size` (now `variants`) method.


sasial-dev (approved):
Thank you!

PR #187, opened 2025-04-16
Fix inaccurate variants value

daimond113:
Previously, the `variants` function's numeric value factored in the possibility of 0 values, but it only needed to be factored in for the size value.


sasial-dev (approved):
Thank you!

PR #188, opened 2025-04-16
Fix regression in recursive type support

daimond113:
Unfortunately, #172 introduced a regression which caused recursive types to overflow the stack. This PR fixes that by using a RefCell to update the type when resolved only once.

### DIFF
--- a/zap/tests/files/recursive.zap
+++ b/zap/tests/files/recursive.zap
@@ -1,0 +1,17 @@
+-- type foo = foo
+
+-- event Simple = {
+--     from: Client,
+--     type: Reliable,
+--     call: SingleSync,
+--     data: foo
+-- }
+
+type bar = struct { baz: bar? }
+
+event Complex = {
+    from: Client,
+    type: Reliable,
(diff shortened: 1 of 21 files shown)


jackhexed (approved):

zap/tests/files/recursive.zap
```diff
@@ -0,0 +1,17 @@
+-- type foo = foo
+
+-- event Simple = {
+--     from: Client,
+--     type: Reliable,
+--     call: SingleSync,
+--     data: foo
+-- }
```
sasial-dev, line 8: Why is this commented out?

```suggestion
type foo = foo

event Simple = {
    from: Client,
    type: Reliable,
    call: SingleSync,
    data: foo
}
```

sasial-dev (approved):
Thanks so much!

zap/tests/files/recursive.zap
daimond113: There currently isn't a way to test for failing tests.

PR #191, opened 2025-04-20
Loosen restrictions on set & list types

daimond113:
Previously, fire_list only accepted arrays of players and fire_set accepted maps of player to true. With this PR, other types are accepted.


sasial-dev (approved):
Thank you!

PR #194, opened 2025-04-25
FIX: Selene flagging missing second argument for assertion

elysianxiv:
Selene flags missing second argument for `assert()` despite it being optional. This PR disables the `incorrect_standard_library_use` lint, which in turn disables the warning.


elysianxiv:
@daimond113 has fixed the assertion issue in #184. I was still running Zap 0.6.19.

There is no longer any need to silence this lint.

PR #195, opened 2025-06-01
Fix TypeScript iteration

daimond113:
roblox-ts doesn't seem to know how to iterate over `[Symbol.iterator]()`, but it does know how to iterate over `IterableFunction` (its own type, also used by e.g. pairs or ipairs). This PR changes the type to allow usage such as:
```typescript
for (const [id, data] of remotes.sendData.iter()) {
```

### DIFF
--- a/zap/src/output/typescript/mod.rs
+++ b/zap/src/output/typescript/mod.rs
@@ -268,12 +268,7 @@ pub trait Output: ConfigProvider {
 	}
 
 	fn push_iter_type(&mut self) {
-		self.push_line("type Iter<T> = () => {");
-		self.indent();
-		self.push_line("[Symbol.iterator](): Iterator<T>,");
-		self.push_line("(): T | undefined");
-		self.dedent();
-		self.push_line("}");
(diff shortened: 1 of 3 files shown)


zap/src/output/typescript/mod.rs
```diff
@@ -268,12 +268,7 @@ pub trait Output: ConfigProvider {
 	}
 
 	fn push_iter_type(&mut self) {
-		self.push_line("type Iter<T> = () => {");
-		self.indent();
-		self.push_line("[Symbol.iterator](): Iterator<T>,");
-		self.push_line("(): T | undefined");
-		self.dedent();
-		self.push_line("}");
+		self.push_line("type Iter<T> = () => IterableFunction<T>");
```
sasial-dev: Let's remove the `Iter` type and instead just do it inline

sasial-dev (changes requested):
Thank you for this! Let's just do it inline as we definitely don't need the `Iter` type now.

sasial-dev (approved):
Thank you!

PR #197, opened 2025-06-23
Remove Ty::Ref where possible

daimond113:
Now, Ty::Ref will only be kept for recursive types. This change will help with future refactors.

Additionally, fixes the following issues:
- possible conflicts of types with the same name in `resolved_tys` due to the lack of scoping
- `unreachable` panic with recursive ORs

### DIFF
--- a/zap/src/parser/convert.rs
+++ b/zap/src/parser/convert.rs
@@ -24,7 +24,7 @@ struct Converter<'src> {
 	tydecls: HashMap<String, SyntaxTyDecl<'src>>,
 	resolved_tys: HashMap<String, Rc<RefCell<Ty<'src>>>>,
 	all_tydecls: HashMap<String, TyDecl<'src>>,
-	current_tydecl_name: Option<&'src str>,
+	current_tydecl: Option<SyntaxTyDecl<'src>>,
 	path: Vec<&'src str>,
 
 	reports: Vec<Report<'src>>,
@@ -44,7 +44,7 @@ impl<'src> Converter<'src> {
 			config,
 			tydecls: HashMap::new(),
 			all_tydecls: HashMap::new(),
-			current_tydecl_name: None,
+			current_tydecl: None,
 			path: Vec::new(),
 			resolved_tys: Default::default(),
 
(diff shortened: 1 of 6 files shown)


zap/src/parser/convert.rs
```diff
@@ -882,6 +929,14 @@ impl<'src> Converter<'src> {
 			}
 
 			let ty = self.ty(syntax_ty);
+			if let Some(current_name) = self.current_tydecl_name {
+				if let Ty::Ref(tydecl) = &ty {
+					if tydecl.path == self.path && tydecl.name == current_name {
+						self.report(Report::AnalyzeOrNestedOptional { span: syntax_ty.span() });
```
sasial-dev: This error doesn't make sense here.

zap/src/parser/convert.rs
```diff
@@ -951,86 +1006,133 @@ impl<'src> Converter<'src> {
...
 
-				match ref_name {
-					ref_name if ref_name == name => Some(*ref_ty),
+				match &*key {
+					ref_name if ref_name == target_path => TyRecursionKind::Unbounded(*ref_ty),
 
-					"boolean" | "Color3" | "Vector3" | "vector" | "AlignedCFrame" | "CFrame" | "unknown" => None,
+					"boolean" | "Color3" | "Vector3" | "vector" | "AlignedCFrame" | "CFrame" | "unknown" => {
+						TyRecursionKind::None
+					}
```
sasial-dev: This doesn't cover all the Ref types... which makes me consider if this block is really needed or if we'll get adverse effects with ones that aren't listed here

zap/src/parser/convert.rs
```diff
@@ -951,86 +1006,133 @@ impl<'src> Converter<'src> {
...
-			SyntaxTyKind::Struct(struct_ty) => self.struct_has_unbounded_ref(name, struct_ty, searched),
+			SyntaxTyKind::Path(path) => {
+				let path = self
+					.path
+					.iter()
+					.copied()
+					.chain(path.iter().map(|i| i.name))
+					.collect::<Vec<_>>()
+					.join(".");
+
```
sasial-dev, line 1059: Does this not need the `searched.contains` block `SyntaxTyKind::Ref` has?

sasial-dev (changes requested):

sasial-dev (approved):
Thank you!

PR #198, opened 2025-06-24
Add bitpacking

daimond113:
Adds bitpacking for booleans - the boolean type and the flag used by optional types to indicate whether it has been serialised.


jackhexed (approved):

zap/src/config.rs
```diff
@@ -431,6 +431,8 @@ impl<'src> Ty<'src> {
 					}
 				}
 
+				let discriminant_numty = NumTy::from_f64(0.0, (or_tys.len() + *optional as usize) as f64);
```
sasial-dev: Although you've added `optional`, I'm worried you are still missing some others from:
```rs
				(used_tys.len()
					+ used_instances.len()
					+ used_variants.len()
					+ prev_unknown_span.is_some() as usize
					+ optional as usize
					- 1) as f64,					
```

zap/src/irgen/mod.rs
```diff
@@ -272,6 +274,109 @@ pub trait Gen {
...
+}
+
+#[derive(Debug)]
+pub enum OutputEntryKind {
+	Stmt(Stmt),
+	Buffer(OutputBuffer),
+}
+
+#[derive(Debug)]
+pub struct OutputEntry(OutputEntryKind);
```
sasial-dev: Why did you wrap `OutputEntryKind` in a `struct`? It shouldn't need to be wrapped.

zap/src/irgen/des.rs
```diff
@@ -18,16 +22,57 @@ impl Gen for Des<'_> {
...
+		self.buf.output()
 	}
 
 	fn get_var_occurrences(&mut self) -> &mut HashMap<String, usize> {
 		self.var_occurrences
 	}
+
+	fn new_scope(&mut self) {
+		let scope_buf = OutputBuffer::new();
+		self.buf.push(scope_buf.clone());
```
sasial-dev, line 42: ~~Why are you pushing here? There's nothing to add to the buffer as it's an empty `Vec`.~~
Does pushing here make sense? It wasn't apparant to me why you did that at a first glance, but there's no real alternative right?

zap/src/irgen/des.rs
```diff
@@ -37,13 +82,35 @@ impl Des<'_> {
...
+					let cond = self.check_bitfield(var.clone(), *bits);
+
+					self.push_stmt(if i == 0 { Stmt::If(cond) } else { Stmt::ElseIf(cond) });
+					self.push_assign(Var::Name(variant_name.clone()), (i as f64).into());
+				}
+				self.push_stmt(Stmt::End);
+				variant_expr
+			}
+			VariantStorageKind::Bit((bits, var)) => {
+				self.check_bitfield(var.clone(), *bits).and(1.0.into()).or(0.0.into())
```
sasial-dev: I wonder if it's more performant to use `bit32.band` here? Or an inline `if` `else` statement?

zap/src/irgen/des.rs
```diff
@@ -37,13 +82,35 @@ impl Des<'_> {
...
+		match storage {
+			VariantStorageKind::Full(numty) => self.readnumty(*numty),
+			VariantStorageKind::Bitpack(variants) => {
+				let (variant_name, variant_expr) = self.add_occurrence("variant");
+				self.push_local(variant_name.clone(), None);
+				for (i, (bits, var)) in variants.iter().enumerate() {
+					let cond = self.check_bitfield(var.clone(), *bits);
+
+					self.push_stmt(if i == 0 { Stmt::If(cond) } else { Stmt::ElseIf(cond) });
+					self.push_assign(Var::Name(variant_name.clone()), (i as f64).into());
```
sasial-dev: We should just assign directly to `into` instead of setting a number and having another if statement for OR/Enums. You could do this by passing a `Vec<Expr>` through and assigning straight to the `i` index of the `Vec`.

zap/src/irgen/des.rs
```diff
@@ -86,10 +154,12 @@ impl Des<'_> {
 		}
 	}
 
-	fn push_or(&mut self, into: Var, tys: &Vec<Ty<'_>>, discriminant_numty: NumTy, optional: bool) {
+	fn push_or(&mut self, into: Var, tys: &Vec<Ty<'_>>, optional: bool) {
 		let (into_ty_i_name, into_ty_i_expr) = self.add_occurrence("ty_i");
+		let storage = self.variant_storage(tys.len() + optional as usize);
```
sasial-dev: Same comment as earlier - let's check this

zap/src/irgen/des.rs
```diff
@@ -172,6 +242,20 @@ impl Des<'_> {
...
+			Var::NameIndex(Var::Name("bit32".into()).into(), "btest".into()).into(),
+			None,
+			vec![Expr::Var(var.into()), Expr::BinaryNum(bits)],
+		)
+	}
+
+	fn readboolean(&mut self) -> Expr {
+		let (bits, var) = self.get_bitpack();
+
+		self.check_bitfield(var, bits)
```
sasial-dev: Was there a reason you swapped the args? You don't have to change it - it just hurt to read!

zap/src/irgen/mod.rs
```diff
@@ -272,6 +274,109 @@ pub trait Gen {
...
+			(1 << existing.0, Var::Name(existing.1.clone()))
+		} else {
+			let (name, _) = self.add_occurrence("bool");
+			self.current_scope().bitpack_budget.push((0, name.clone()));
+			(1, Var::Name(name))
+		}
+	}
+
+	fn variant_storage(&mut self, amount: usize) -> VariantStorageKind {
+		// 0 is variant 1, 1 is variant 2
```
sasial-dev: This comment doesn't make sense. We also need to handle `amount <= 1`

zap/src/irgen/mod.rs
```diff
@@ -272,6 +274,109 @@ pub trait Gen {
...
+		if amount <= 2 {
+			VariantStorageKind::Bit(self.get_bitpack())
+		} else if self.current_scope().remaining_bitpack_budget() as usize >= amount {
```
sasial-dev, line 302: Just confirming, the rationale behind `>= amount` is performance right?

zap/src/irgen/mod.rs
```diff
@@ -469,6 +575,7 @@ impl Display for Expr {
 
 			Self::Var(var) => write!(f, "{var}"),
 			Self::Num(num) => write!(f, "{num}"),
+			Self::BinaryNum(num) => write!(f, "0b{num:016b}"),
```
sasial-dev: I feel like we should instead of padding to a `u16`, we should pad to the nearest type (u8, u16, u32)

As in, this should really be padded to a `u8`:
```lua
			value[1] = bit32.btest(bool_5, 0b0000000000000001)
			value[2] = bit32.btest(bool_5, 0b0000000000000010)
			value[3] = bit32.btest(bool_5, 0b0000000000000100)
			value[4] = bit32.btest(bool_5, 0b0000000000001000)
			value[5] = bit32.btest(bool_5, 0b0000000000010000)
			value[6] = bit32.btest(bool_5, 0b0000000000100000)
			value[7] = bit32.btest(bool_5, 0b0000000001000000)
			value[8] = bit32.btest(bool_5, 0b0000000010000000)
```

zap/src/irgen/ser.rs
```diff
@@ -96,6 +165,17 @@ impl Ser<'_> {
 			)),
 		);
 
+		let storage = self.variant_storage(
+			tys.iter()
+				.map(|ty| match ty.primitive_ty() {
+					PrimitiveTy::Enum(Enum::Unit(variants)) => variants.len(),
+					PrimitiveTy::Enum(Enum::Tagged { variants, .. }) => variants.len(),
+					_ => 1,
```
sasial-dev, line 174: What about optional/unknown?

zap/src/irgen/ser.rs
```diff
@@ -186,19 +266,38 @@ impl Ser<'_> {
...
+		);
+	}
+
+	fn push_bool<F: FnOnce(&mut Self)>(&mut self, cond_expr: Expr, cb: F) {
+		let (bits, var) = self.get_bitpack();
+		self.push_stmt(Stmt::If(cond_expr));
+		self.set_bitfield(var, bits);
+		cb(self);
+		self.push_stmt(Stmt::End);
+	}
```
sasial-dev: Don't require a conditional expression/callback here... just have `push_bool` have `self.get_bitpack` and `self.set_bitfield`.

sasial-dev (changes requested):

sasial-dev (approved):
Thank you so much!

PR #201, opened 2025-07-02
Separate is empty and length serde

daimond113:
Previously, to serialise all variants of `set { u8 }`/`map { [u8]: u8 }`, a u16 was used for the length. This was wasteful, as we only needed 1 more number to serialise whether the map is empty or not. Since now bitpacking is an option, the length is serialised as `[0, 255]` and 1 is added/subtracted on each side, and the emptiness is a bitflag.


sasial-dev (approved):
Thank you!

PR #203, opened 2025-07-03
Add precise string kinds

daimond113:
Adds 2 new types: `string.utf8` and `string.binary`. `string` behaves the same as `string.binary`, although it is now deprecated & emits a warning when used - users are expected to explicitly choose `string.binary`.

`string.utf8` additionally checks for the string being valid UTF-8. The upper bound is now in characters - it gets multiplied by 4 because UTF-8 characters can take up to 4 bytes. This change is to prevent users from sending, e.g. 3-byte strings which may be too little for a UTF-8 string.
Since UTF-8 is a variable-width encoding, the length is always serialised, even if the range is exact.

### DIFF
--- a/zap/src/config.rs
+++ b/zap/src/config.rs
@@ -324,7 +324,7 @@ impl<'src> Ty<'src> {
 			Self::Num(numty, ..) => (numty.size(), Some(numty.size())),
 
 			Self::Str(utf8, len) => {
-				if !*utf8 && let Some(exact) = len.exact() {
+				if !utf8 && let Some(exact) = len.exact() {
 					(exact as usize, Some(exact as usize))
 				} else {
 					let (len_numty, ..) = len.numty();
--- a/zap/src/config.rs
+++ b/zap/src/config.rs
@@ -272,7 +272,7 @@ impl Display for TyDecl<'_> {
 #[derive(Debug, Clone)]
 pub enum Ty<'src> {
 	Num(NumTy, Range),
-	Str(Range),
+	Str(bool, Range),
 	Buf(Range),
 	Vector(Box<Ty<'src>>, Box<Ty<'src>>, Option<Box<Ty<'src>>>),
 	Arr(Box<Ty<'src>>, Range),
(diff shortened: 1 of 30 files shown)


zap/src/config.rs
```diff
@@ -323,8 +323,8 @@ impl<'src> Ty<'src> {
 		match self {
 			Self::Num(numty, ..) => (numty.size(), Some(numty.size())),
 
-			Self::Str(len) => {
-				if let Some(exact) = len.exact() {
+			Self::Str(utf8, len) => {
+				if !*utf8 && let Some(exact) = len.exact() {
```
sasial-dev: Do you need to own the bool here? Couldn't you just do `!utf8`?

zap/src/parser/reports.rs
```diff
@@ -251,6 +257,7 @@ impl Report<'_> {
 			Self::AnalyzeOrNestedOptional { .. } => "3021",
 			Self::AnalyzeRecursiveOr { .. } => "3022",
 			Self::AnalyzeConflictingExport { .. } => "3023",
+			Self::AnalyzeNoStringDataKind { .. } => "3024",
```
sasial-dev: I'm wondering if we should create a new kind of error for deprications (`4001`). Depends on if we choose to PR the instance syntax changes before the next release.

zap/src/irgen/des.rs
```diff
@@ -238,7 +238,11 @@ impl Des<'_> {
 						self.push_range_check(len_expr.clone(), *range);
 					}
 
-					self.push_assign(into, self.readstring(len_expr.clone()));
+					self.push_assign(into.clone(), self.readstring(len_expr.clone()));
+
+					if *utf8 {
+						self.push_utf8_check(Expr::Var(into.into()));
+					}
```
sasial-dev, line 245: I think this is the only place (unlike array) where we assign to `into` before validating... I think that is okay though?

sasial-dev (changes requested):

sasial-dev (approved):
Thank you!

PR #204, opened 2025-07-04
Add path-like class specifiers for instances

daimond113:
Brings new syntax `Instance.Player` to replace `Instance (Player)`. The new syntax looks much nicer when combined with other syntax: `Instance.Player[1..2]`, `Instance (Player)[1..2]`, `Instance.Player?`, `Instance (Player)?` as well as being more consistent.


sasial-dev (approved):
Thank you!

PR #206, opened 2025-07-20
Fix bitpacking mismatch

daimond113:
Fixes a mismatch in the bits used by bitpacking when dealing with 2 variants.


sasial-dev (approved):
Thank you!

PR #209, opened 2025-10-05
Fix garbage packets with function calls

daimond113:
Fixes the event id being written to the buffer before validating that there aren't too many waiting calls.


sasial-dev (approved):
Thanks for finding it!

PR #210, opened 2025-10-08
Make reliable default

Ezzenix:
Add type_default option similar to existing call_default option

### DIFF
--- a/zap/src/config.rs
+++ b/zap/src/config.rs
@@ -250,6 +250,12 @@ pub enum EvType {
 	Unreliable(bool),
 }
 
+impl Default for EvType {
+    fn default() -> Self {
+        EvType::Reliable
+    }
+}
+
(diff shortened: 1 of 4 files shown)


sasial-dev:
Please fix CI!

Ezzenix:
Sorry i have not used rust before

zap/src/parser/reports.rs
```diff
@@ -35,6 +35,10 @@ pub enum Report<'src> {
 
 	AnalyzeEmptyEvDecls,
 
+	AnalyzeMissingEvDeclType {
+		span: Span,
+	},
```
sasial-dev: nit: all instances of `AnalyzeMissingEvDeclType` should be in order of error codes like the rest of them.

sasial-dev (changes requested):
As per @nezuo's great suggestion - let's change this to just be Reliable by default instead.

sasial-dev (approved):
Thanks!

PR #212, opened 2025-11-27
Add `include_profile_labels` option to include microprofiler labels

nezuo:
Feel free to suggest alternative names for the option.

Here are the places labels were added:

- For callbacks:
```lua
local function profiledCallback(label, callback)
	return function(...)
		debug.profilebegin(label)
		callback(...)
		debug.profileend()
	end
end

setCallback = function(callback: (message: (string)) -> ())
    callback = profiledCallback("EventName Callback", callback)
```
- When events/functions are fired/called: `EventName Fire`/`EventName FireAll`/`FunctionName Call`
- In OnClientEvents/OnServerEvents: `Zap Reliable OnClientEvent`/`Zap UnreliableEventName OnClientEvent`
- A label for each event/functions's deserialization: `EventName Deserialize`
- A label for `sendEvents()`: `Zap Send Events`

### DIFF
--- a/docs/.vitepress/components/Editor.vue
+++ b/docs/.vitepress/components/Editor.vue
@@ -2,13 +2,13 @@
 	<MonacoEditor
 		:value="props.modelValue"
 		@update:value="(val: string) => $emit('update:modelValue', val)"
-		:language="lang ?? 'zapConfig'"  
+		:language="lang ?? 'zapConfig'"
 		:theme="`${props.isCodeBlock ? 'codeblock' : 'tab'}-${isDark ? 'dark' : 'light'}`"
 		@beforeMount="beforeMount"
 		@mount="(editor) => $emit('mounted', editor)"
 		:options="EDITOR_OPTIONS"
-	/>	
-</template>	
+	/>
+</template>
 
 <script setup lang="ts">
 import MonacoEditor from "@guolao/vue-monaco-editor";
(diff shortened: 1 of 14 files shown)


docs/.vitepress/components/Editor.vue
```diff
@@ -2,13 +2,13 @@
...
-		:language="lang ?? 'zapConfig'"  
+		:language="lang ?? 'zapConfig'"
 		:theme="`${props.isCodeBlock ? 'codeblock' : 'tab'}-${isDark ? 'dark' : 'light'}`"
 		@beforeMount="beforeMount"
 		@mount="(editor) => $emit('mounted', editor)"
 		:options="EDITOR_OPTIONS"
-	/>	
-</template>	
+	/>
+</template>
```
sasial-dev, line 11: Can you fix the diff here please?

sasial-dev (changes requested):
Can you please add tests? They'll point out an issue I noticed as to where you're setting a global variable for example.

Image

docs/.vitepress/components/Editor.vue
nezuo: What's the issue here? It was just my editor removing trailing whitespace.

nezuo:
Didn't realize we had tests, I'll take a look.

docs/.vitepress/components/Editor.vue
sasial-dev: Been a fan of not outgrowing the scope of the PR, but I agree it's fine to keep in.

nezuo:
Callbacks are no longer wrapped in `profiledCallback` and now there are just profiles around them wherever they are called.

I also added the `display_path` methods to `FnDecl` and `EvDecl` so we can show `Namespace.Event` instead of just `Event`.

sasial-dev (approved):
Thank you!

PR #217, opened 2026-02-12
Docs: Updated deprecated Instance and String syntax

ConfidentLetter:


ConfidentLetter:
`options.md` on line 140 was not updated as I interpreted it to just be an indicator of length and not a proper type declaration.

sasial-dev (approved):
Thank you!

PR #218, opened 2026-03-15
Feat: Allow requiring zap from edit mode (for Storyboards)

EstebenR:
Studio will return true for RunService:IsServer() in edit mode. For testing components using storyboards, any component that requires the client network module will error.

This fix makes it so Zap is allowed to be required while in edit mode even if IsServer returns true

### DIFF
--- a/zap/src/output/luau/client.rs
+++ b/zap/src/output/luau/client.rs
@@ -1457,7 +1457,7 @@ impl<'src> ClientOutput<'src> {
 	}
 
 	pub fn push_check_server(&mut self) {
-		self.push_line("if RunService:IsServer() then");
+		self.push_line("if RunService:IsServer() and not RunService:IsEdit() then");
 		self.indent();
 		self.push_line("error(\"Cannot use the client module on the server!\")");
 		self.dedent();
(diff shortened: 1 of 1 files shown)


Ezzenix:
IsEdit has plugin security it will error when playing.

EstebenR:
> IsEdit has plugin security it will error when playing.

Error how? I currently have this change live in my games and there's no issue

Ezzenix:
Then you are not requiring from the server in your live game. Since RunService:IsServer() is false the RunService:IsEdit() never runs for you.

EstebenR:
Working as intended then? If ClientNetwork were to be required from the server in live it should error anyways, shouldn't it?
I guess we could wrap it in a pcall in that case so that the error is descriptive... What do you suggest?

Ezzenix:
Calling IsEdit() outside of a plugin will give this error
"The current thread cannot call 'IsEdit' (lacking capability Plugin)"

I think what you want is
```
if RunService:IsServer() and RunService:IsRunning() then
	error("Cannot use the client module on the server!")
end
```

EstebenR:
Yep you're right. I've tested it and it works for storyboards :D

sasial-dev:
Realised I forgot to hit approve - but thank you!

PR #223, opened 2026-05-02
Dont send after PlayerRemoving

### LINKED ISSUE #216: [BUG] `Player` is memory leaked if `Fire` is called after `PlayerRemoving`
## Describe the bug

After the `PlayerRemoving` event fires for a player, if any calls are made to `Fire` or another dispatcher, zap adds that Player to `player_map` again on the server. This is then never cleaned up and persists forever.

## Reproduction

```zap
event Test = {
    from: Server,
    type: Reliable,
    call: ManyAsync,
    data: (),
}
```
Playground [link](https://zap.redblox.dev/playground?code=ZXZlbnQgVGVzdCA9IHsNCiAgICBmcm9tOiBTZXJ2ZXIsDQogICAgdHlwZTogUmVsaWFibGUsDQogICAgY2FsbDogTWFueUFzeW5jLA0KICAgIGRhdGE6ICgpLA0KfQ%3D%3D)

```lua
local Players = game:GetService("Players")
local network = require("./network/server")

Players.PlayerRemoving:Connect(function(player)
	task.wait(5)
	network.Test.Fire(player)
	print("Player will remain in player_map forever!")
end)
```

## Expected behavior
Although this event has been fired in error (after player has left), zap should protect against this or issue a warning when this occurs. As it stands, an event firing seconds after the player leaves silently causes a memory leak.

Ezzenix:
Fix #216 by checking if player is still in player_map and also adding to player_map on PlayerAdded.

### DIFF
--- a/zap/src/output/luau/server.rs
+++ b/zap/src/output/luau/server.rs
@@ -518,6 +518,8 @@ impl<'src> ServerOutput<'src> {
 			self.push_line(&format!("task.spawn(function(player_2, call_id_2, {args})"));
 			self.indent();
 
+			self.push_line("if player_map[player_2] == nil then return end");
+
 			self.push_line(&format!(
 				"local {rets_string} = reliable_events[{server_id}](player_2, {args})"
 			));
@@ -830,6 +832,8 @@ impl<'src> ServerOutput<'src> {
 		self.push(")\n");
 		self.indent();
 
+		self.push_line("if player_map[player] == nil then return end");
+
 		if self.config.include_profile_labels {
 			self.push_line(&format!("debug.profilebegin(\"{} Fire\")", ev.display_path()));
 		}
@@ -1085,6 +1089,7 @@ impl<'src> ServerOutput<'src> {
 				self.push_line("local buff, used, inst = outgoing_buff, outgoing_used, outgoing_inst");
 				self.push_line(&format!("for _, player in {list} do"));
 				self.indent();
+				self.push_line("if player_map[player] == nil then continue end");
 				self.push_line("load_player(player)");
 				self.push_line("alloc(used)");
 				self.push_line("buffer.copy(outgoing_buff, outgoing_apos, buff, 0, used)");
@@ -1102,6 +1107,7 @@ impl<'src> ServerOutput<'src> {
 				}
 				self.push_line(&format!("for _, player in {list} do"));
 				self.indent();
+				self.push_line("if player_map[player] == nil then continue end");
 				if ordered {
 					self.push_unreliable_order_id("player", ev.id);
 				}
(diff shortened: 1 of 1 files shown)


zap/src/output/luau/server.rs
```diff
@@ -1604,6 +1612,18 @@ impl<'src> ServerOutput<'src> {
...
+		self.push_line("load_player(player)");
+		self.push_line("player_map[player] = save()");
+		self.dedent();
+		self.push_line("end");
+		self.push_line("Players.PlayerAdded:Connect(function(player)");
+		self.indent();
+		self.push_line("load_player(player)");
+		self.push_line("player_map[player] = save()");
+		self.dedent();
+		self.push_line("end)");
```
sasial-dev: This is an unnecessary addition

sasial-dev (changes requested):

zap/src/output/luau/server.rs
Ezzenix: If they are not added to player_map on join they will only be added after the server receives an event from the player, so zap would drop their requests before that. But i just noticed an issue, other scripts PlayerAdded will run before zaps so firing inside PlayerAdded would no longer work.

zap/src/output/luau/server.rs
```diff
@@ -518,6 +518,8 @@ impl<'src> ServerOutput<'src> {
 			self.push_line(&format!("task.spawn(function(player_2, call_id_2, {args})"));
 			self.indent();
 
+			self.push_line("if player_map[player_2] == nil and player_2.Parent == nil then return end");
```
sasial-dev: As a performance optimisation, we should check if the player exists *before* spawning a task.

zap/src/output/luau/server.rs
```diff
@@ -830,6 +832,8 @@ impl<'src> ServerOutput<'src> {
 		self.push(")\n");
 		self.indent();
 
+		self.push_line(&format!("if player_map[{player}] == nil and {player}.Parent == nil then return end"));
```
sasial-dev: This will return in places it shouldn't, as we save to the player map (which could be for the first time) about 25 lines below.

zap/src/output/luau/server.rs
```diff
@@ -1102,6 +1107,7 @@ impl<'src> ServerOutput<'src> {
 				}
 				self.push_line(&format!("for _, player in {list} do"));
 				self.indent();
+				self.push_line("if player_map[player] == nil and player.Parent == nil then continue end");
```
sasial-dev, line 1112: We don't use `player_map` for unreliables.

sasial-dev (changes requested):

